PR #11, opened 2016-11-15
Workflows

avirshup:
MIGRATED FROM PRIVATE REPO

This PR will collect my work to get a few workflows organized and running before leaving on vacation :)


avirshup:
@amagod - this is reorganized and up and running. If you want to give it a shot, run:

./build-nodes.sh to build both the Docker image and the node definitions.
./download_and_clean_3aid.sh to run a sample workflow (note it uses a custom cache location)

PR #12, opened 2016-11-15
Wires up dummy API for client.

dionjwa:
@justinmc 

The server now has `POST /pdb_convert' where the body of the post request if the PDB file data. 

*For now* it just returns the same PDB file, you can then take that for the client. I'll be wiring up Aaron's actual workflow after lunch, but this should unblock you for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,19 +2,13 @@
 
 Predictive molecular modeling applications based on the [Molecular Design Toolkit](https://github.com/Autodesk/molecular-design-toolkit) framework. (Early development, all features are subject to change)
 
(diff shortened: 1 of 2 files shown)


dionjwa:
@justinmc I've just updated this branch. I got the `/pdb_convert` API end point working correctly. You can test it with 

    curl --data-binary "@5e8b.pdb" http://localhost:4000/pdb_convert

(Note its `--data-binary` not `--data` otherwise newlines get removed.

The path to the output pdb is returned, you can then use a GET on the url or whatever you need.

PR #15, opened 2016-11-18
(WIP) Update molecule-3d by parsing pdb

justinmc:
Currently we're using a modified version of molecule-3d that takes in a PDB instead of json.  This PR will add a pdb->json converter to the code so that we can use the latest release of molecule-3d like any other application.

### DIFF
--- a/client/.eslintrc
+++ b/client/.eslintrc
@@ -5,6 +5,8 @@
     "document": true,
     "fetch": true,
     "FileReader": true,
+    "it": true,
+    "describe": true,
   },
   "rules": {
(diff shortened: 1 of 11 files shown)


justinmc:
This should be done for the most part, but I'm unable to run the full app right now and see changes in my javascript for some reason...  Will check on this and then test it out before merging.

justinmc:
Closing this since we've moved away from 3dmol and over to the Autodesk molviewer.  Note that this PR contains work on a nice tested PDB parser that could be useful though!

PR #16, opened 2016-12-02
 New Design

justinmc:
This will transform our old prototype UI into the [new design](https://app.zeplin.io/project.html#pid=582f3a02541d3d2b506e1a59&dashboard) and general functionality.  Only modifying the frontend for now with mock endpoints, will probably save the backend for another PR.


justinmc:
I think I'm going to merge this at this point.  I've implemented everything major except for the "thank you" page.  Everything is still ugly and needs another pass to make it match the designs, but it's in a place that should make that easy as I collaborate with Malte.

justinmc:
@dionjwa Heads up that this is going in.  Let me know if you want a chance to code review stuff like this in the future, sorry for merging while you're offline :)

PR #25, opened 2016-12-15
Thank you page

justinmc:
This PR includes an ugly version of the thank you page, as well as different states for the page for like canceled and errored out.


justinmc:
@dionjwa Heads up that this is going in.  UI only.

PR #29, opened 2016-12-15
Results Pane

justinmc:
This creates the results pane (ugly for now) and gets most of the functionality on there working as best as it can.


justinmc:
@dionjwa Another UI only PR going in.

PR #30, opened 2016-12-15
(WIP) Server and Frontend Connection Work

justinmc:
Starting to create routes needed by frontend.

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -1,0 +1,8 @@
+# Workflow Web Server
+This workflow serves the API for the web app frontend and also communicates with the workflow execution server.
+
+## API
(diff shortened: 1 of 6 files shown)


justinmc:
@dionjwa Let me know what you think of this before I continue on this path.  I implemented a fake "get workflow" route that would return a workflow from the database (not implemented yet) given its id.  Then I just added some basic express boilerplate from the express generator that I usually use for logging, error handling, etc.

justinmc:
Whoops I totally didn't think I merged this, I must have thought it was another PR last night.  I will continue this work in another PR 😕

PR #31, opened 2016-12-16
Error and Running Pages

justinmc:
This PR copies over Malte's designs for the Error and Running pages.


justinmc:
Pending Malte's approval on the looks, then will merge.

PR #33, opened 2016-12-16
Working docker computes and routes

dionjwa:
- The server uses a persistent cache for storing emails and workflow states
- There is a dummy workflow (that still runs in a docker container) that just copies inputs to outputs. This verifies running the container, exiting, status codes, stdout/err etc.
- Routes for getting this data added (see README).
- Emails are not yet done.


server/workflow.js
```diff
@@ -1,56 +1,507 @@
...
+  /*
+    busboy doesn't wait until piped file streams emit a close event,
+    and this means that the data isn't actually written before the
+    busboy.finish event is fired, so any reading of written files
+    will fail. By using promises we can listen to the 'close' event
+    ensuring that reading those files will work.
+   */
+  var promises = [];
+  var email = null;
+  var error = null;
```
justinmc: `error` is never assigned a value if I'm not mistaken

server/workflow.js
```diff
@@ -1,56 +1,507 @@
...
+  });
+  return Hash(hashes);
+}
+
+/**
+ * [executeWorkflow description]
+ * @param  {[type]} workflowDir [description]
+ * @return {[type]}             [description]
+ */
+function executeWorkflow(workflowId) //Promise<Bool>
```
justinmc: I would move all of these helper functions into another file or files.  I usually have a `utils/` directory with utils files that contain stuff like this.  Moving them into their own module helps enforce that they don't sneakily share state or closure variables from this file, and it makes them testable.

justinmc: Also I think you mean to put that "Promise<Bool>" comment into your jsdoc comment above.

server/workflow.js
```diff
@@ -1,56 +1,507 @@
...
+            }
+          }, (err) => {
+            console.log(err);
+          });
+      });
+    }, (err) => {
+      console.error(err);
+    });
+};
+reattachExistingWorkflowContainers();
```
justinmc: Are you intending to call this when this routes file is required?

dionjwa:
I'm merging this for now, but those are good suggestions (because this is Javascript, ugh, I am too used to a proper language where you actually have classes and proper static functions). I'll add these suggestions next week.

justinmc:
Sounds good haha, thanks for taking a look on Saturday.

PR #34, opened 2016-12-16
Linting and Compliance

justinmc:
This adds eslint to the project and fixes any linting errors that existed.


justinmc:
@dionjwa I think I fixed all the lint errors in workflow.js, let me know what you think.  Then I think this is good to merge.

PR #37, opened 2016-12-16
Connect Front/Backends

justinmc:
~~Branched off of: https://github.com/Autodesk/molecular-design-applications/pull/34~~

This PR connects the backend and frontend and normalizes all inconsistencies.


justinmc:
@dionjwa I had planned to wait till you got back to merge this so I could get your review, but I'm going to need to merge it to keep PR complexity down.  We'll have to sit down at some point to go over the changes and come up with plans for going forward.  The main changes to what you wrote were the distinction between workflows and runs, and the storing of workflow and run data in Redis (for now).  Otherwise I tried to keep it as similar as possible.

Giving it a last review myself and then merging.

PR #43, opened 2017-01-03
App Runs Via Docker

justinmc:
With this PR, you should now be able to run the app with:

    $ docker-compose build && docker-compose up

You should see the app chrome with a "404" on the page at the home root "/".  This is expected for now.  Another PR (https://github.com/Autodesk/molecular-design-applications/pull/37) is tying together the front and backends to make it actually work.

This should also work when deployed, hopefully!


justinmc:
@PeterRJones This PR should no longer be needed, since all of its changes just went in with https://github.com/Autodesk/molecular-design-applications/pull/37.  The master branch should be able to run via docker now (works for me locally at least).

PR #49, opened 2017-01-06
Cancel

justinmc:
This gets the cancel button working with the real backend.


justinmc:
Merging this without review while @dionjwa is out.

PR #54, opened 2017-01-06
Homepage

justinmc:
This PR will add the homepage (built by Malte) to the app in a way that puts us in a good position for small bundle sizes and future dynamic content.


justinmc:
@dionjwa Another one you might want to take a look at when you're back.  The main interesting thing here is just code splitting and including the homepage and the actual simulation app as different asset bundles.

PR #59, opened 2017-01-11
Emails

justinmc:
Gets emails working with a real design from Malte.  Waiting for approval from Malte and then should be good to go.


justinmc:
Issue: https://github.com/Autodesk/molecular-design-applications/issues/52

justinmc:
Malte gave a 👍

PR #63, opened 2017-01-13
Status pages and results UI

justinmc:
This updates the status pages with Malte's UI, and it modifies the run UI to use real data from the VDE run.


justinmc:
@dionjwa You can review/take a look at this before I merge it.  It's mostly copying in Malte's UI stuff, but I also did a little bit of data reorganization with workflows and views on the frontend.  I also fixed a small 404 bug in the server.

client/public/js/records/workflow_record.js
```diff
@@ -1,30 +1,18 @@
...
-  runId: null,
+  run: new RunRecord(),
   runs: 0,
-  status: statusConstants.IDLE,
   title: '',
-  uploadError: null,
-  uploadPending: false,
   views: 0,
   workflowNodes: new IList(),
 });
```
dionjwa: Where is this record saved? Is this going in the cache?

dionjwa:
Justin, you are going to need to walk me through this application for me to do meaningful code reviews.

client/public/js/records/workflow_record.js
justinmc: Sorry I forgot to reply to this.  Let me know if this still isn't clear after our talk earlier.  Basically these records are saved in stores on the client only, but they include a bunch of data that will usually also need to be saved in the backend (in Redis for now).

PR #74, opened 2017-01-17
PDB By Id

justinmc:
This PR makes "get pdb by id" work by directly downloading the pdb from rcsb.  Currently branched off of https://github.com/Autodesk/molecular-design-applications/pull/63 so do not merge until that goes in.

### DIFF
--- a/client/public/js/actions.js
+++ b/client/public/js/actions.js
@@ -174,23 +174,18 @@ export function submitPdbId(pdbId) {
       type: actionConstants.SUBMIT_PDB_ID,
     });
 
-    let pdbUrl;
-    apiUtils.getPdbById(pdbId).then((responsePdbUrl) => {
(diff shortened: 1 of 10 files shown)


justinmc:
A slight concern I have with this is that it is not able to be used offline.  You can get around this by uploading a pdb instead of using this narrow feature though, so the app overall is not broken when running locally.  To get around this, we could write a proxy endpoint that could be mocked, but I don't think that actually improves the situation.

justinmc:
We will need to parse the pdb at upload/fetch time anyway in the mvp workflow, so I think we should keep the pdb_by_id endpoint.

justinmc:
I'm going to put this PR up for review right now with basic PDB_by_id functionality and save any MVP workflow stuff for another PR.

PR #82, opened 2017-01-19
Input pdb processing

justinmc:
Currently branched off of https://github.com/Autodesk/molecular-design-applications/pull/74.

This PR will set us up to do PDB processing in a docker container when the user selects an input pdb.


README.md
```diff
@@ -47,11 +47,11 @@ Returns the indicated run with its workflow populated.
...
 Runs the indicated workflow.  Requires workflowId, email, and pdbUrl.
 
-##### GET /structure/pdb_by_id/:pdbId
-Returns a url to the pdb file represented by the given pdbId.
+##### GET /structure/pdb_by_id?pdbId&workflowId
+Returns the pdb data and a url to the pdb file represented by the given pdbId.  For some workflow ids, the pdb will be processed before being returned, and a `data` parameter will also be included with additional data.
 
 ##### PUT /structure/upload
-Uploads the given pdb file to the server and returns a public URL to it.
+Uploads the given pdb file to the server and returns a public URL to it.  Sends formdata with a `workflowId` and a `file`.
```
justinmc: Note these 2 api changes, which are implemented below.

client/public/js/utils/api_utils.js
```diff
@@ -26,15 +27,17 @@ const apiUtils = {
...
+        resolve(`${API_URL}${res.data.pdbUrl}`)
       ).catch(reject);
     });
   },
 
-  getPdbById(pdbId) {
-    return axios.get(`${API_URL}/v1/structure/pdb_by_id/${pdbId}`).then(res =>
+  getPdbById(pdbId, workflowId = '1') {
+    const query = querystring.stringify({ pdbId, workflowId });
+    return axios.get(`${API_URL}/v1/structure/pdb_by_id?${query}`).then(res =>
```
justinmc: Passing workflowId with both the `getPdbById` and `upload` api calls.

server/routes/structure.js
```diff
@@ -1,27 +1,94 @@
...
+            return ioUtils.stringToHashFile(pdb, 'public/structures').then(
+              processedFilename =>
+                res.send({
+                  pdbUrl: `/structures/${processedFilename}`,
+                  pdb,
+                  data,
+                })
+            ).catch(next);
+          }
+        ).catch(next);
```
justinmc: For both the `upload` and `pdb_by_id` endpoints, we now call out to `workflowUtils.processInput` to do the processing.  The processed pdb data and url are returned in both cases.

server/utils/redis.js
```diff
@@ -1,8 +1,20 @@
...
+// Seed data
+dbUtils.seed(redis, dbConstants.REDIS_WORKFLOWS, {
+  id: '0',
+  title: 'VDE',
+}).catch(console.error.bind(console));
+dbUtils.seed(redis, dbConstants.REDIS_WORKFLOWS, {
+  id: '1',
+  title: 'Preparing the outer ligand structure',
+}).catch(console.error.bind(console));
+
```
justinmc: @dionjwa Seed data. I used this to make sure the code distinguishes between workflows that need input processing and those that don't.

server/utils/workflow_utils.js
```diff
@@ -0,0 +1,28 @@
...
+
+        // Handles workflows that don't require input processing (like VDE)
+        default:
+          return resolve({ pdb: '', data: {} });
+      }
+    });
+  },
+};
+
+module.exports = workflowUtils;
```
justinmc: @avirshup This is where any calls to run the actual input processing in a docker container would be (somehow reusing our main workflow run docker stuff).  Note the switch statement for hardcoding which workflows need processing and which don't.  We need a better solution than that at some point.

README.md
dionjwa: I don't think that the server should be responsible for downloading a PDB only to return the PDB to the client. It adds complexity and is unnecessary, as well as slower (two trips instead of one). Couldn't the client simply download the PDB file directly?

If the PDB needs to be modified, then an API should expressly do that.

client/public/js/actions.js
```diff
@@ -168,29 +168,24 @@ export function upload(file) {
...
       dispatch({
         type: actionConstants.FETCHED_PDB_BY_ID,
-        error: err.message,
-      });
-    });
+        err: err.message,
+      })
+    );
   };
 }
```
dionjwa: If the client downloaded PDB files when given a PDB id, then these actions can be simplified, and only deal with PDB data (once the PDB is downloaded by the client).

client/public/js/reducers/workflow.js
```diff
@@ -11,59 +10,47 @@ function workflow(state = initialState, action) {
       const workflowsDifferent = action.workflowId !== state.id;
       if (workflowsDifferent) {
         return new WorkflowRecord({
-          run: new RunRecord({
-            fetching: true,
-            fetchingError: false,
-          }),
+          fetching: true,
+          fetchingError: false,
         });
```
dionjwa: Would it make more sense to have an `error` field, and if null, there is no error, otherwise it contains the error (maybe stringified) itself?

server/routes/structure.js
```diff
@@ -1,27 +1,94 @@
...
+    if (fieldname === 'workflowId') {
+      workflowId = val;
+    }
+  });
+
   busboy.on('file', (fieldname, file) => {
     ioUtils.streamToHashFile(file, 'public/structures').then((filename) => {
-      res.send({
-        path: `/structures/${filename}`,
+      fs.readFile(`public/structures/${filename}`, 'utf8', (err, inputPdb) => {
```
dionjwa: Should we promisify these via Bluebird.promisify?

server/utils/io_utils.js
```diff
@@ -31,37 +32,118 @@ const ioUtils = {
         hash.end();
         resolve(hash.read());
       });
-      readableStream.on('error', reject);
+      readableStream.on('error', (err) => {
+        reject(err);
+      });
```
justinmc: I agree.  I think I was doing some debugging here and needed a function to put the `debugger` in or something like that, and I forgot to change it back.

server/utils/io_utils.js
```diff
@@ -31,37 +32,118 @@ const ioUtils = {
...
+                ).catch(reject);
+              });
+              writeableStream.on('error', reject);
+              return fs.createReadStream(tempFilepath).pipe(writeableStream);
+            }
+
+            return ioUtils.deleteFile(tempFilepath).then(() =>
+              resolve(filename)
+            ).catch(reject);
+          });
```
justinmc: Good call, that's much cleaner.

server/routes/structure.js
justinmc: I've never used promisify but it looks good to me, it definitely cleans up the code.  I'm pushing a commit that promisifies `fs` in this file, and then I'll try to move that way for everything else in the future.

client/public/js/reducers/workflow.js
justinmc: Actually yeah for `fetchingError`, it looks like I am setting it to the error string when there is an error, but I should set it to `null` otherwise so it doesn't look like it's a boolean.

server/utils/workflow_utils.js
avirshup: @justinmc Cool, this certainly works for now (of course, we'll want the actual ligand selector in there :) ).

Without knowing anything about front-end architecture, from my perspective I'd love for these to be as workflow-like as possible, so that a workflow developer can easily define what needs to happen with a given PDB file.

client/public/js/actions.js
justinmc: Let's talk about these upload/process/run endpoints tomorrow after standup with @avirshup.  I like the idea of more atomic endpoints, but I need to draw this all out on the whiteboard and make sure everyone is on the same page with how this flow will work.

server/utils/workflow_utils.js
justinmc: Cool sounds good.  I'm trying to keep everything as generic as possible but still end up hardcoding things like this switch statement.  Maybe the solution is to have a field for `inputProcessingImage` and if present then we'll call it with the input, and if not we'll skip processing.

justinmc:
Merging after talking to @dionjwa.  Will address the upload/process/run endpoint stuff in a separate PR.

PR #83, opened 2017-01-21
Test Endpoint

justinmc:
Issue: https://github.com/Autodesk/molecular-design-applications/issues/73

Sample successful call:

```
➜  server git:(master) ✗ http GET http://localhost:4001/v1/test
HTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Connection: keep-alive
Content-Length: 26
Content-Type: application/json; charset=utf-8
Date: Sat, 21 Jan 2017 00:44:03 GMT
ETag: W/"1a-Q1vgZEVxrmJJRvQ8k0qwbQ"
X-Powered-By: Express

{
    "runId": "test-S1sA07xPl"
}
```

### DIFF
--- a/server/routes/test.js
+++ b/server/routes/test.js
@@ -1,4 +1,4 @@
-"use strict";
+'use strict';
 
 const express = require('express');
 const shortid = require('shortid');
@@ -14,14 +14,14 @@ const router = new express.Router();
  * Tries running a test run. Does everything the same as /run except no emails.
  * Returns success if submitting the run succeeds (doesn't care if the run
  * itself actually succeeds or not).
- * TODO: change above so that workflow job is submitted and results are verified 
- * to have succeeded before this route returns. 
+ * TODO: change above so that workflow job is submitted and results are verified
+ * to have succeeded before this route returns.
  */
 router.get('/', (req, res, next) => {
   const workflowId = seedData[0].id;
-  const runId      = `test-${shortid.generate()}`;
-  var   resReturn  = { runId, success: true };
-  const email      = 'test@autodesk.com';
+  const runId = `test-${shortid.generate()}`;
+  const resReturn = { runId, success: true };
+  const email = 'test@autodesk.com';
   const inputPdbUrl = 'https://files.rcsb.org/download/3aid.pdb';
 
   const runPromise = redis.hset(dbConstants.REDIS_RUNS, runId, JSON.stringify({
@@ -37,6 +37,7 @@ router.get('/', (req, res, next) => {
   );
 
   return Promise.all( [runPromise, statePromise] ).then(() => {
+
     runUtils.executeWorkflow(runId, inputPdbUrl);
 
     const runUrl = `${process.env.FRONTEND_URL}/workflow/${workflowId}/${runId}`;
--- a/server/routes/test.js
+++ b/server/routes/test.js
@@ -1,3 +1,5 @@
+"use strict";
+
 const express = require('express');
 const shortid = require('shortid');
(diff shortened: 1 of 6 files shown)


PeterRJones:
@justinmc if you add a `success` or `result` field to the json that would be preferred. then the validation logic would be the same as for other apps. Something like this:

`{
    "runId": "test-S1sA07xPl",
   "success": true
}`

for the semantics of your api, if that matters, since the call is async, this `success` would merely mean the submission was successful.  

the deployment logic considers `/test` successful if: http result == 200 and either `success` or `result` == true.

if adding this goes against other plans or api design, let's sync up Monday for a fix.

PeterRJones:
@justinmc Please review. This works locally. 
One concern and one thing for future TODO.

Concern is that this will actually create a workflow run for each call. This will be called once during deployment and every 5 minutes by monitor. Are you okay with the volume of calls due to monitor ?

TODO for later: This route should, in future update, block until the entire call has succeeded or failed and return the appropriate result. That the desired end state.

justinmc:
@dionjwa @avirshup What do you guys think about this test endpoint spinning off a VDE run every 5 minutes?  Is there a simple way that achieves the same test coverage but doesn't use so many resources?

dionjwa:
If the molecule is very small, the compute time is also very short, in the order of a few seconds. I think this is ok. If the health check doesn't actually check integration with computes, it's not a very useful test, even if this does consume resources. 

@justinmc I have more comprehensive tests and some of this duplicated in my upcoming PR. I'll make two versions of the test route: one testing all, and other for the health check (only running one).

server/routes/test.js
```diff
@@ -1,3 +1,5 @@
+"use strict";
```
justinmc: This makes me realize that we should be running node with `--use_strict`

server/routes/test.js
```diff
@@ -12,11 +14,14 @@ const router = new express.Router();
...
  * itself actually succeeds or not).
+ * TODO: change above so that workflow job is submitted and results are verified 
+ * to have succeeded before this route returns. 
  */
 router.get('/', (req, res, next) => {
   const workflowId = seedData[0].id;
-  const runId = `test-${shortid.generate()}`;
-  const email = 'test@autodesk.com';
+  const runId      = `test-${shortid.generate()}`;
+  var   resReturn  = { runId, success: true };
```
justinmc: This `var`, and maybe some other changes here, will break our linter rules.  This isn't really enforced anywhere so maybe it doesn't really matter, but you can run eslint with `node_modules/.bin/eslint routes/test.js` from the server directory to see what's in violation.

justinmc:
@dionjwa Sounds good, so @PeterRJones I guess we will go ahead with this PR and then make incremental improvements as we go?

server/routes/test.js
PeterRJones: Got my Sublimetext JSHint lint settings closer to `eslint` but could not get them to exactly match.  Here are the final lint errors remaining - should be okay for now to let these persist:

```
molecular-design-applications/server/routes/test.js
   1:1   error  'use strict' is unnecessary inside of modules                            strict
   5:33  error  Unable to resolve path to module 'molecular-design-applications-shared'  import/no-unresolved
  36:35  error  Missing trailing comma                                                   comma-dangle
  39:21  error  There should be no spaces inside this paren                              space-in-parens
  39:50  error  There should be no spaces inside this paren                              space-in-parens
  39:63  error  Block must not be padded by blank lines                                  padded-blocks
  ```

PR #84, opened 2017-01-24
Docker image fixes

dionjwa:
- For local development, uses a local CCC server.
- All docker computes are done in CCC (with the npm CCC client lib).
- Refactored root *js files into a separate folder ('main') to allow mounting server files for faster local development.
- Added tests and test routes for testing workflow0.
- Partial migration to storing less uploaded files, and moving them to CCC.
- Bunyan logging added. Fluent not yet hooked up, but little configuration to do so.
- Errors are JSONable.

NB workflow/0 still has some issues that I cannot identify. The format of the final output is not clear to me.


server/package.json
```diff
@@ -14,23 +14,30 @@
...
     "dotenv": "^4.0.0",
     "express": "^4.14.0",
     "fs-extended": "^0.2.1",
+    "http-proxy": "^1.16.2",
     "molecular-design-applications-shared": "../shared",
     "morgan": "^1.7.0",
     "mustache": "^2.3.0",
     "promise-redis": "0.0.5",
     "redis": "^2.6.3",
+    "request": "^2.79.0",
```
justinmc: I've been using axios elsewhere but I don't have much preference if you like request.

server/package.json
```diff
@@ -14,23 +14,30 @@
...
     "sendgrid": "^4.7.1",
     "shortid": "^2.2.6",
-    "validator": "^6.2.0"
+    "validator": "^6.2.0",
+    "ws": "^1.1.1"
```
justinmc: Is this used anywhere?  Maybe I missed it, hard to ctrl-f.

server/routes/run.js
```diff
@@ -13,22 +14,27 @@ const router = new express.Router();
...
     if (!runString) {
       const error = new Error(`Run "${req.params.runId}" not found`);
       error.status = 404;
+      console.error(error);
       return next(error);
     }
 
     const run = JSON.parse(runString);
+    if (run.outputPdbPath && run.outputPdbPath.indexOf('ccc:9000') > -1) {
+      run.outputPdbPath = run.outputPdbPath.replace('ccc:9000', 'localhost:9000');
```
justinmc: Should we replace this before it goes into the database, or is there a reason to have it in other places?

server/routes/run.js
```diff
@@ -50,43 +56,45 @@ router.post('/', (req, res, next) => {
...
-
-  return Promise.all([emailPromise, runPromise, statePromise]).then(() => {
-    runUtils.executeWorkflow(runId, req.body.pdbUrl);
+  var getInputPromise = null;
+  if (req.body.pdbData) {
+    getInputPromise = Promise.resolve(req.body.pdbData);
+  } else if (req.body.pdbUrl) {
+    var pdbUrl = req.body.pdbUrl;
+    if (!pdbUrl.startsWith('http')) {
+      pdbUrl = `http://localhost:${process.env.PORT}${pdbUrl}`;
```
justinmc: I made an environment variable `process.env.URL` for things like this, in case it's not running on localhost.

server/routes/structure.js
```diff
@@ -1,6 +1,8 @@
 const Busboy = require('busboy');
 const express = require('express');
 const ioUtils = require('../utils/io_utils');
+const appConstants = require('../constants/app_constants');
+log.warn('appConstants=' + JSON.stringify(appConstants));
```
justinmc: Did you mean to log this?

server/routes/structure.js
```diff
@@ -14,14 +16,24 @@ router.get('/pdb_by_id/:pdbId', (req, res, next) => {
...
   const busboy = new Busboy({
     headers: req.headers,
   });
 
+  busboy.on('field', (fieldname, file) => {
+    log.trace({api:'upload', event:'field', field:fieldname});
+  });
+  busboy.on('error', (error) => {
+    log.error({message:'on busboy error', error:error});
+    next(error);
```
justinmc: Good call, I totally missed handling these errors.

server/test/test_utils.js
```diff
@@ -0,0 +1,125 @@
...
+      }
+    })
+    .catch(err => {
+      log.error(err);
+      return {success:false, error:JSON.stringify(err)};
+    });
+  },
+};
+
+module.exports = test_utils;
```
justinmc: Should we get these tests working with a test framework like mocha or something?

server/utils/io_utils.js
```diff
@@ -1,8 +1,29 @@
...
 const appRoot = require('app-root-path');
 const crypto = require('crypto');
-const fs = require('fs');
+const fs = require('fs-extended');
 const path = require('path');
+const shortId = require('shortid');
+
+
+/**
+ * Hash the contents of the given stream
```
justinmc: "steam" should read "file"

server/utils/io_utils.js
```diff
@@ -1,8 +1,29 @@
...
+ * @param readableStream {Stream}
```
justinmc: And this should be a filepath

server/package.json
```diff
@@ -14,23 +14,30 @@
...
+    "bluebird-retry": "^0.10.1",
     "body-parser": "^1.15.2",
+    "bunyan": "^1.8.5",
+    "bunyan-format": "^0.2.1",
     "busboy": "^0.2.13",
+    "cloud-compute-cannon-client": "^0.1.39",
     "cookie-parser": "^1.4.3",
     "cors": "^2.8.1",
     "debug": "~2.2.0",
     "dockerode": "^2.3.1",
```
justinmc: Can we remove this since we're using CCC now?

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
+      });
+      return redis.hset(
+        dbConstants.REDIS_RUNS, runId, JSON.stringify(updatedRun)
+      );
+    })
+    .catch(console.error.bind(console))
+    .then(ignored => {
+      runUtils.sendEmailsWorkflowEnded(runId);
+    });
+  },
```
justinmc: I worry about it being possible for some runs to error out and not have the database updated to reflect the error, so the user would always see the "thank you" page.  Nothing to do with this PR though, just a thought I've had while working on this "job finished" code.

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
+
+        const runUrl = `${process.env.FRONTEND_URL}/workflow/${workflowId}/${runId}`;
+        emailUtils.send(
+          email,
+          'Your Workflow is Running',
+          './views/email_thanks.ms',
+          { runUrl }
+        );
+        //Record the jobId with all the other data
+        const emailPromise = redis.sadd(dbConstants.REDIS_WORKFLOW_EMAIL_SET, email);
```
justinmc: Can we get rid of this now?  I'm using the run redis hash to store the email, not sure if this set is used anywhere anymore.

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
+
+        const runPayload = {
+          id: jobId,
+          workflowId,
+          email: email,
+          // params: params,
```
justinmc: What is params?

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
+          created: Date.now(),
+        };
+        log.debug(JSON.stringify(runPayload).substr(0, 300));
+
+        const runPromise = redis.hset(dbConstants.REDIS_RUNS, jobId, JSON.stringify(runPayload));
+        const statePromise = runUtils.setRunStatus(jobId, statusConstants.RUNNING);
```
justinmc: We can do both of these in one call to redis.hset.  We might as well set the status of the run here instead of calling to setRunStatus for a second hset.

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
-                container.remove({ force: 1 }, (errRemove) => {
-                  if (errRemove) {
-                    console.error(`Run=${runId} Error removing container=${containerData.Id} err]${errRemove}`);
-                  }
-                });
-                runUtils.sendEmailsWorkflowEnded(runId);
+  monitorRun(runId) {
+    log.debug('Monitoring run ' + runId);
+    runUtils.waitOnJob(runId)
+      .then(result => {
```
justinmc: Will this have to be something more complicated like a queue in the long term?  If I'm reading this right and this promise may have to wait for 24 hours before resolving.

server/utils/run_utils.js
```diff
@@ -158,131 +51,167 @@ const runUtils = {
...
+   * @return {[type]}        [description]
    */
-  executeWorkflow(runId, inputPdbUrl) {
-    console.log(`executeWorkflow ${runId}`);
-    runUtils.getDockerImage(WORKFLOW_DOCKER_IMAGE)
-      .then(() => {
-        console.log('got image');
-        return new Promise((resolve, reject) => {
-          console.log(`executeWorkflow runId=${runId}`);
+  executeWorkflow0(params) {
```
justinmc: What is this method for?

justinmc:
Also, you should set up a linter in your editor if you can.  There is a .eslintrc file in the project that should give us a common styleguide.  I want to add it to a CI check at some point.

server/utils/run_utils.js
dionjwa: Yes, there are still edge cases here not accounted for. In that case, the important thing is to log all errors. 

In this case, the user should probably be presented with a page saying something along the lines of "Your workflow is finished, but there was a problem. Contact XXX". And those errors should probably be bubbled up to us.

server/utils/run_utils.js
dionjwa: Yes, this is *not* a long term solution for running long term workflows. A better approach is a queue, but I would want to rethink the entire system before committing to queues for now.

server/utils/run_utils.js
dionjwa: We have more than one workflow:
```/* When we have more than one workflow, we'll switch on the workflow Id */
    const workflowPromise = runUtils.executeWorkflow0(params);```

Currently, I don't have the details for the other workflow Aaron is working on, I only have this first one. But the sequence of steps is the same, so this will reduce duplicate code when I add the second workflow.

File: docker-compose.override.yml [hunk 17]
```diff
@@ -9,12 +9,75 @@ services:
...
+    image: kibana:4.4.2
+    ports:
+      - "5601:5601"
+    environment:
+      ELASTICSEARCH_URL: "http://elasticsearch:9200"
+    #Pretty sure I don't care about listening to you ever
+    logging:
+      driver: none
+    links:
+      - elasticsearch
```
dionjwa: It should require zero effort from your part, just a `docker-compose up`, but if you hit problems, just let me know so I can make the improvements. The goal is that `docker-compose up` will always provide a zero install (apart from docker) way to build and run an entire stack, specifically for development, hence mounting all the src directories in the `docker-compose.override.yml` file.

server/bin/www
```diff
@@ -82,10 +116,36 @@ function onError(error) {
...
+    const redis = require('../utils/redis');
+    const dbConstants = require('../constants/db_constants');
+    return redis.hset(dbConstants.REDIS_WORKFLOWS, 0, JSON.stringify({ id: 0, title: 'VDE' }));
+}
+
+/**
+ * Proxy requests to port 9000 to the CCC server
+ */
+const httpProxy = require('http-proxy');
+httpProxy.createProxyServer({target:'http://ccc:9000'}).listen(9000);
```
dionjwa: This is only required for the local development case. In the production case, the CCC server is not part of the stack, so there aren't any problems using a proper CCC url, e.g. `10.34.45.34:9000`. However, in the local development case, the CCC server is linked to the main server, but this causes difficulties when passing result URLs back to the client. I would need to write logic to detect if the main server is a development docker-compose stack, and modify the URLs within our main server (e.g. from `foo/bar/result.json` to `localhost:9000/foo/bar/result.json`. By using a proxy, it removes the need for development-only specific logic. In production, this proxy does nothing, but also uses no resources, so I did not put it in a ENV var.

server/package.json
dionjwa: Er, this is the CCC client. This is a javascript library compiled from CCC (thanks to Haxe) that gives you an object with methods and returns promises. It's a lot easier to use than calling your own REST calls and parsing the data. The client library handles all that for you.

server/package.json
dionjwa: I have no preference. If axios is better, let's use that.

server/routes/structure.js
dionjwa: No, good catch. I'll remove that.

server/package.json
dionjwa: It was used in the CCC client, but now that package correctly installs it itself. I'll remove.

server/test/test_utils.js
dionjwa: Sure, if you think it's worth it. When the test function is just calling methods that return promises, I haven't seen much value add by the bigger frameworks. And I've often been working against a health check call that covers the core functions. 

But I would like to see the benefits of whatever test framework you recommend.

server/utils/io_utils.js
dionjwa: Yes, good catch.

server/utils/io_utils.js
dionjwa: Done.

server/routes/run.js
dionjwa: I don't see this anywhere. Can you be more specific? This is the problem with Javascript as the main language, it's all just string searching to find code.

dionjwa: I'll make a change to CCC to automatically append the localhost URL prefix when run in a docker-compose stack locally.

PR #85, opened 2017-01-24
FIX: jenkins build

dionjwa:
- Add auto-rebuilding for the client with the dev docker-compose config.
 - Remove obsolete client Dockerfile.


justinmc:
Looks good to me 👍

PR #96, opened 2017-01-27
Workflow1 first step.

dionjwa:
- Workflow1, step 1 (process pdb for ligand selection).
 - Test included.
 - Tested with public CCC as well.


server/routes/structure.js
```diff
@@ -6,11 +6,13 @@ const fs = Promise.promisifyAll(require('fs'));
 const ioUtils = require('../utils/io_utils');
 const workflowUtils = require('../utils/workflow_utils');
 const appConstants = require('../constants/app_constants');
+const shortid = require('shortid');
 
 const router = new express.Router();
 
 const RCSB_URL = 'https://files.rcsb.org/download';
 
+/* JUSTIN: Is this obsolete? */
```
justinmc: I'll be working on branch `workflow1-frontend`.

server/routes/structure.js
```diff
@@ -48,53 +50,47 @@ router.get('/pdb_by_id', (req, res, next) => {
...
-  busboy.on('file', (fieldname, file) => {
-    ioUtils.streamToHashFile(file, 'public/structures').then((filename) => {
-      fs.readFileAsync(`public/structures/${filename}`, 'utf8').then((err, inputPdb) => {
-        if (!workflowId) {
-          return next(new Error('Needs a valid workflow id.'));
-        }
+  const tmpFileName = `/tmp/_temp_executeWorkflow1Step0_${shortid.generate()}`;
+  const cleanup = () => {
+    try {
+      fs.deleteFileSync(tmpFileName);
```
justinmc: I'm always afraid of using synchronous stuff due to performance hits in Node, but I'm not sure how much it matters for us in reality.

server/utils/workflow_utils.js
```diff
@@ -1,4 +1,110 @@
...
+
+    return ccc.run(jobJson, cccInput)
+      .then(jobResult => {
+        if (jobResult.exitCode == 0) {
+          log.info(jobResult);
+          var baseUrl = jobResult.outputsBaseUrl;
+          var result = {
+            success: true,
+            "prepJson": baseUrl + 'prep.json',
+            "prepPdb": baseUrl + 'prep.pdb'
```
justinmc: Did you get your linter working?  I think I saw a few things in this PR that will break it, like this mix of double and single quotes.  Not a big deal though, we have to clean everything up at some point later.

server/bin/www
```diff
@@ -70,6 +70,7 @@ Promise.all(seedData.map(workflow =>
   server.listen(port);
   server.on('error', onError);
   server.on('listening', onListening(server.address()));
+  server.timeout = 8 * 50 * 1000; // 8 minutes
```
justinmc: Should that be 60 instead of 50?

server/routes/run.js
```diff
@@ -47,51 +47,30 @@ router.get('/:runId', (req, res, next) => {
  * Start a run
  */
 router.post('/', (req, res, next) => {
-  if (!req.body.workflowId && req.body.workflowId !== 0) {
+  var workflowId = req.body.workflowId;
+  var email = req.body.email;
+  var inputs = req.body.inputs;
+  log.info({email:email});
+  log.info({workflowId:workflowId});
```
justinmc: These vars and objects will break the linter, but we can fix that later.

server/routes/structure.js
```diff
@@ -8,57 +8,65 @@ const shortid = require('shortid');
...
 
 /**
- * First step in workflow1: selecting a ligand.
- * Test with:
- * curl -F file=@`pwd`/server/test/1bna.pdb localhost:4000/v1/structure/executeWorkflow1Step0
- * @param  {[type]}   '/executeWorkflow1Step0' [description]
- * @param  {Function} (req,                    res,          next)         [description]
- * @param  {[type]}   'utf8').then((err,       inputPdb      [description]
- * @return {[type]}                            {"prepJson": "URL", "prepPdb": "URL"}
+ * https://docs.google.com/presentation/d/1qP-8fPpsgtJnZOlg96ySwPACZvGlxT1jIIgjBECoDAE/edit#slide=id.g1c36f8ea4a_0_0
```
justinmc: This is in a public Github repo and the Google Dock is shared by link.  That sounds fine to me, but just pointing it out.

server/routes/structure.js
```diff
@@ -8,57 +8,65 @@ const shortid = require('shortid');
...
+ * }
+ * @return {"success": true, "outputs": {"filename":"url"}, "jobResult": <CCC Job for debugging>}
  */
-router.post('/executeWorkflow1Step0', (req, res, next) => {
-  const busboy = new Busboy({
-    headers: req.headers,
-  });
+router.post('/executeWorkflow0Step0', (req, res, next) => {
+  var inputs = req.body.inputs;
+  if (!inputs) {
```
justinmc: What if inputs is an empty array?

server/routes/test.js
```diff
@@ -6,19 +6,50 @@ const testUtils = require('../test/test_utils');
...
+router.get('/workflow0', (req, res) => {
+  testUtils.runTestWorkflowVDE()
+    .then((result) => {
+      if (result.success) {
+        res.send(result);
+      } else {
+        res.sendStatus(500).send(result);
+      }
+    }, (err) => {
+      res.sendStatus(500).send(JSON.stringify({test:'failed', error:JSON.stringify(err)}));
```
justinmc: You shouldn't have to manually stringify this json (unless I'm missing something about how you're using this output).

server/test/test_utils.js
```diff
@@ -1,22 +1,29 @@
...
 const retry = require('bluebird-retry');
+const request = require('request-promise');
+const workflowUtils = require('../utils/workflow_utils');
+
 const statusConstants = require('molecular-design-applications-shared').statusConstants;
 
 const test_utils = {
   runAllTests() {
-    return Promise.all([test_utils.runTest1(), test_utils.runTest2(), test_utils.runTest_executeWorkflow1Step0()])
+    return Promise.all([test_utils.runTestWorkflowVDE(), test_utils.runTestWorkflowVDE()])
```
justinmc: Did you mean to call `runTestWorkflowVDE` twice?

server/test/test_utils.js
```diff
@@ -27,124 +34,147 @@ const test_utils = {
...
-              var getResult = JSON.parse(body);
-              const status = getResult.status;
-              if (status === statusConstants.RUNNING) {
-                log.trace('Rejecting because body.status=' + status);
-                reject(getResult);
-              } else {
-                resolve(getResult);
+    return Promise.resolve(true)
+      //Step 1
+      .then(ignored => {
```
justinmc: You can just `() => {` instead of creating the `ignored` parameter, and the linter will complain about that too.  Actually it's probably clearest just to `return request.post...` directly and then `.then` off of that without using an unneeded `Promise.resolve`.

server/test/test_utils.js
```diff
@@ -27,124 +34,147 @@ const test_utils = {
...
+          name: 'input.pdb',
+          value: fs.readFileSync('test/3aid.pdb', {encoding:'utf8'})
+        }
+      ]
+    };
+
+    var port = process.env.PORT;
+    const url = `http://localhost:${port}/v1/structure/executeWorkflow1Step0`;
+    //Step 1
+    return Promise.resolve(true)
```
justinmc: See my comment above about the extra Promise.resolve

server/utils/workflow_utils.js
```diff
@@ -153,4 +180,6 @@ const workflowUtils = {
   },
 };
 
+workflowUtils.ccc = ccc;
```
justinmc: I'm worried about exposing ccc here leading to bad architecture...  Would a `getStatus` method be better?  It seems like that's all ccc is used for outside of this file right now.

server/routes/structure.js
```diff
@@ -8,57 +8,65 @@ const shortid = require('shortid');
...
-  busboy.on('file', (fieldname, file) => {
-    const writeStream = fs.createWriteStream(tmpFileName);
-    writeStream.on('finish', () => {
-      workflowUtils.executeWorkflow1Step0(fs.createReadStream(tmpFileName))
-        .then((jobResult) => {
-          if (!jobResult.success) {
-            const error = new Error('Failed to execute processing');
-            error.result = jobResult;
-            handleError(error);
-            return;
```
justinmc: My frontend is expecting to see an error status when the processing doesn't succeed, so I need something like this if statement.

server/bin/www
dionjwa: Whoops yes. Fixed.

server/routes/structure.js
```diff
@@ -1,101 +1,72 @@
 const Busboy = require('busboy');
 const Promise = require('bluebird');
-const axios = require('axios');
 const express = require('express');
 const fs = Promise.promisifyAll(require('fs'));
-const ioUtils = require('../utils/io_utils');
 const workflowUtils = require('../utils/workflow_utils');
-const appConstants = require('../constants/app_constants');
+const shortid = require('shortid');
```
justinmc: This isn't used in this file

server/routes/test.js
```diff
@@ -6,21 +6,55 @@ const testUtils = require('../test/test_utils');
...
+  testUtils.runTestWorkflowQMMM()
+    .then((result) => {
+      if (result.success) {
+        res.send(result);
+      } else {
+        res.status(500).send(result);
+      }
+    })
+    .catch((err) => {
+      res.status(500).send(JSON.stringify({success:false, error:JSON.stringify(err)}));
```
justinmc: I commented this before but I think I messed it up: Is this JSON.stringify necessary?

server/test/test_utils.js
justinmc: Which seems to have disappeared.  It said you could return `request.post` directly and get rid of the extra Promise.resolve

server/routes/structure.js
dionjwa: Fixed

server/utils/ccc_utils.js
```diff
@@ -0,0 +1,36 @@
...
+  retry(
+    function() {
+      return ccc.status()
+        .then(status => {
+          return status;
+        });
+    },
+    {max_tries: 50, interval:1000}
+  )
+  .then(ignored => {
```
justinmc: You can do `() => {` and get rid of the `ignored` parameter.

server/routes/test.js
dionjwa: Will `JSON.stringify()` get called on the error object, or just a form of `toString()`? If it's the latter, then it is needed to get the full error stack, otherwise you often get only partial error info.

server/utils/ccc_utils.js
dionjwa: Fixed.

server/test/test_utils.js
dionjwa: I missed that comment. I add that in there to keep the bodies of long promise chains consistently formatted. Then promise blocks can be easily edited by moving them around, rather than formatting the top one.

server/utils/ccc_utils.js
```diff
@@ -0,0 +1,36 @@
+const Promise = require('bluebird');
+const retry = require('bluebird-retry');
+
+const CCCC = require('cloud-compute-cannon-client');
+const ccc = CCCC.connect(process.env["CCC"]);
+
+var cccPromise =
+  retry(
```
justinmc: Is this going to execute only once immediately when the file is included?  I think you should paste this stuff into the `promise` method declaration instead of using a closure variable like this.  I could be wrong but it looks like a bug at first glance...

server/test/test_utils.js
dionjwa: Anyway, removed.

server/routes/run.js
dionjwa: Fixed.

server/routes/test.js
justinmc: I think it's calling `JSON.stringify` since it looks the same to me.  Here is the results of `return res.send(new Error('faillll'));` and `return res.send(JSON.stringify(new Error('faillll')));` in that order:


screen shot 2017-02-06 at 11 05 25 am

dionjwa: Ok, good. I'll remove the last stringify

server/utils/ccc_utils.js
dionjwa: Yes, the whole point of the `require` system is that modules only get called once. Then the module is cached, unless it's purposefully deleted. Repeated execution blocks need to go in exposed functions.

justinmc: So when this file is required, this whole block from line 8 to 28 is immediately executed.  Whenever you call `cccUtils.promise`, it's just returning the existing result of that original execution and never running it again.  Is that what you wanted?

dionjwa: Yes. CCC may go down, and in that case, retrying should be the tactic. But here, we're waiting until we have a response from the CCC daemon at least *once* before allowing any CCC operations. It also allows deleting all jobs in the dev mode, when the server is constantly restarting due to edits.

justinmc: Ok cool.  I'm flipping my review to 👍

PR #105, opened 2017-01-31
Workflow1 frontend

justinmc:
This will get the frontend of the app working with the api changes in the workflow1 input processing PR https://github.com/Autodesk/molecular-design-applications/pull/96.

### DIFF
--- a/client/public/css/button.scss
+++ b/client/public/css/button.scss
@@ -62,6 +62,11 @@
     box-shadow: 0 1px 0px rgba(0, 0, 0, 0.075);
     -webkit-transition: 0.2s; 
     transition: 0.2s;
+
+    &.disabled {
+      border: 1px solid rgba(0, 0, 0, 0.15);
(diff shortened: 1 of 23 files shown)


justinmc:
@dionjwa Review needed please!  Here are the main changes:

  - Use of RCSB api to fetch pdbs by id moved to the frontend
  - Upload endpoint removed
  - Frontend calls `executeWorkflow1Step0` endpoint for workflow 1 whenever input is given, whether by selecting a pdb file or entering a pdb id.  So workflow 1 is hardcoded into the frontend which isn't good, but we'll make this generic in the front and back ends later.
  - Frontend deals in raw pdb strings instead of urls now.
  - Run submission is broken!  The run endpoint wants a url, but we don't necessarily have one now (see previous point).
  - Tiny server change: error status is returned when the execution of `executeWorkflow1Step0` fails.

PR #114, opened 2017-02-06
Add content

avirshup:
@justinmc @MalteTinnus Here's a draft of material for the home page, as well as app names for the gallery


server/utils/seed_data.js
```diff
@@ -1,25 +1,48 @@
...
+        bgColor: '#292E60',
+        bgIndex: 1,
+        color: '#2FE695',
+        comingSoon: true,
+        creatorImage: '/assets/logo2.png',
+        description: "Predict a small molecule's perceived color and its UV-Viz spectra using TD-DFT",
+        viewCount: 0,
+    },
+    {
+        id: '1',
```
justinmc: @avirshup Can you increment these ids so that they're all unique?  Otherwise I think this PR looks good.

Also all of these fields are just a big guess at what kinds of things we want to customize for the appearance of each workflow.  If you or @MalteTinnus need to customize something else then let me know and we can program that in.

avirshup:
@justinmc - this is up-to-date and ready to merge, can you take a look?

avirshup:
Screenshot of current state:
image

justinmc:
Looks awesome with all the real content there 👍

avirshup:
Real-ish content, anyway :)

PR #117, opened 2017-02-08
Step0 API Change Frontend Adjustment

justinmc:
This PR gets the frontend working again with step0.  For workflow0, you should be able to submit a run.  For workflow1, ligand selection is not working due to no molviewer integration yet.

### DIFF
--- a/client/public/js/components/workflow.jsx
+++ b/client/public/js/components/workflow.jsx
@@ -5,6 +5,7 @@ import Status from '../components/status';
 import View from '../components/view';
 import WorkflowRecord from '../records/workflow_record';
 import WorkflowSteps from '../components/workflow_steps';
+import ioUtils from '../utils/io_utils';
 import selectionConstants from '../constants/selection_constants';
 
 require('../../css/workflow.scss');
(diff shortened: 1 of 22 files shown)


client/public/js/components/workflow.jsx
```diff
@@ -14,20 +15,20 @@ function Workflow(props) {
...
+    selectedModelData = ioUtils.getInputPdb(props.workflow.run.inputs);
   } else if (props.selection.type ===
     selectionConstants.WORKFLOW_NODE_RESULTS) {
-    if (props.morph === 1) {
-      selectedModelData = props.workflow.run.outputPdb;
-    } else {
-      selectedModelData = props.workflow.run.inputPdb;
-    }
+    const modelDatas = props.workflow.run.inputs
+      .concat(props.workflow.runs.outputs)
```
dionjwa: Just a type? `modelDatas` => `modelData`?

client/public/js/utils/rcsb_api_utils.js
```diff
@@ -8,8 +8,8 @@ const rcsbApiUtils = {
     return axios.get(pdbUrl).then(res => ({
       pdbUrl,
       pdb: res.data,
-    })).catch((err) => {
-      throw err;
+    })).catch(() => {
+      throw new Error('Couldn\'t find a pdb with that id, please try again.');
     });
```
dionjwa: So is this what the linter insists? Use a backslash for single quotes, rather than double quotes and avoid backslashes?

justinmc: It seems to be fine with either way

client/public/js/components/workflow.jsx
justinmc: I was trying to indicate that it's an array of model data

PR #125, opened 2017-02-10
Workflow 1 End-to-End

justinmc:
This will get workflow1 working end-to-end.  Yes molviewer ligand selection, no smiles probably for now.


avirshup:
This looked mostly good when merging with master. Only issues were:
1) Morphing didn't seem to work, and
2) Selecting a ligand in step 2 didn't affect the viewer

justinmc:
The two issues above should be fixed...  But not 100% sure about morphing.

Another issue I just noticed is that Ligand Selection isn't checked on the results page:

screen shot 2017-02-13 at 7 17 43 pm

I thought I had a quick fix by feeding in the ligand selection data, but if I do that, it behaves weirdly and tries to visualize the ligand during morph.

The real fix will require a bit of refactoring.  In fact, I'm not 100% sure that this idea of the results page being just an extension of the pre-run page.  It might not work at all as workflows get more complex.

avirshup:
I've done a test merge, and for the sake of getting a functional production deployment, I'm going to go ahead and merge this into master

PR #129, opened 2017-02-11
Server fixes

dionjwa:
- Extra logging
 - ensure Error objects are thrown (not strings)
 - Ensure workflowId is a string

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -26,67 +26,66 @@ services:
       # - ./client/dist:/app/client/dist
     command: ["nodemon", "-e", "js,json,www", "-w", "./", "-w", "./bin/www", "bin/www"]
     environment:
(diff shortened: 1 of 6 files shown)


docker-compose.override.yml
```diff
@@ -26,66 +26,67 @@ services:
...
+  #   image: kibana:4.4.2
+  #   ports:
+  #     - "5601:5601"
+  #   environment:
+  #     ELASTICSEARCH_URL: "http://elasticsearch:9200"
+  #   #Pretty sure I don't care about listening to you ever
+  #   logging:
+  #     driver: none
+  #   links:
+  #     - elasticsearch
```
justinmc: Why is all this stuff commented out?  I don't fully understand the docker setup anyway but just curious :)

PR #131, opened 2017-02-11
Updates to terminology, correct output fields for VDE

avirshup:


avirshup:
Changing various strings and names for the frontend

client/public/js/components/status_results.jsx
```diff
@@ -43,7 +43,7 @@ class StatusResults extends React.Component {
         <div className="stats">
           <div className="stat">
             <div className="fontHeader line stat-title">
-              <div>OUTPUT ENERGY</div>
+              <div>VERTICAL DETACHMENT ENERGY</div>
```
justinmc: This will be hardcoded, so it will show up for workflow1 also.  But if we're just worrying about VDE for now then this is ok, because there's no super easy to do this dynamically.

client/public/js/components/status_email.jsx
```diff
@@ -40,8 +40,8 @@ class StatusEmail extends React.Component {
 
   render() {
     const sendText = this.props.email ?
-      `We will send an email to you at ${this.props.email} once your workflow is completed.` :
-      'We will send you an email once your workflow is completed';
+      `We'll send you an email at ${this.props.email} when you start this calculation, and another when it's done.` : // eslint-disable-line max-len
+      'We\'ll send you an email at when you start this calculation, and another when it\'s done.';
```
justinmc: The linter might complain about not using `&#39;` for that single-quote, or for line length on the line above, so watch out for that.  If not then no worries.

client/public/js/components/status_results.jsx
avirshup: Yeah, I was worried about that. I guess we're trying to get VDE done today, so hardcoding is OK, but will it be possible to have different displays next week? Not sure I can think of a label here that would apply to both minimization and VDE

justinmc: The first thing that comes to my head is to add a field in the workflow seed data for "outputDataLabel" or something like that.  That wouldn't be super elegant but is probably the best solution, and yeah it would be fairly quick to do.

PR #133, opened 2017-02-13
Textcleanup

avirshup:


avirshup:
@justinmc Ready for review

client/public/js/components/home_page.jsx
```diff
@@ -210,15 +207,16 @@ class HomePage extends React.Component {
...
                 <p>
                   It's still early days and there's lots to be done. If you're interested in
                     developing your own deployable applications - get in contact with us. We're
                     looking for good use cases that will have material impact on molecular design,
                     and can work with you to help give your research impact for a wide audience.
                 </p>
               </div>
+              <div className="col-xs-12 col-sm-4">
+
+              </div>
```
justinmc: You could also make these two `col-sm-6` divs if you'd prefer it to be full-width divided in half, instead of 2/3 width and the last third empty like right now.  I think it looks fine either way though, just letting you know in case you wanted to do that.

client/public/js/components/workflow_card.jsx
```diff
@@ -46,19 +46,19 @@ function WorkflowCard(props) {
           backgroundImage: `url(${bgImage})`,
         }}
       />
-      <img
+{/*      <img
         src={`${process.env.API_URL}${props.creatorImage}`}
         alt="card logo"
         className="cardLogo"
-      />
+      />*/}
```
justinmc: Are we leaving these out until we have better images?

avirshup: I think we're ready to put these in, just probably not before tomorrow - I just commented out everything that wasn't hooked up.

client/public/js/components/home_page.jsx
avirshup: Oh, that's much easier - will do

client/public/js/components/workflow_card.jsx
justinmc: Ok sounds good.

justinmc:
@avirshup Feel free to merge whenever you're ready!

PR #135, opened 2017-02-13
Email URLs

justinmc:
This fixes the emails linking to "undefined".  From https://github.com/Autodesk/molecular-design-applications/issues/76.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -10,6 +10,7 @@ services:
     environment:
       CCC: "ccc.bionano.autodesk.com:9000"
       NODE_ENV: "production"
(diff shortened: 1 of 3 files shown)


avirshup:
Looks good on submission:
image
And good when done:
image

docker-compose.prod.yml
```diff
@@ -10,6 +10,7 @@ services:
     environment:
       CCC: "ccc.bionano.autodesk.com:9000"
       NODE_ENV: "production"
+      FRONTEND_URL: "http://molsim.bionano.autodesk.com/"
     logging:
```
dionjwa: This env var is created, but it's not used anywhere?

PR #136, opened 2017-02-13
Dynamic name on results pane

justinmc:
Previously, the results page has 1 field that is hard coded to show "OUTPUT ENERGY".  This PR will make that title dynamic.

### DIFF
--- a/client/public/js/components/status.jsx
+++ b/client/public/js/components/status.jsx
@@ -91,7 +91,7 @@ function Status(props) {
     );
   } else if (props.selection.type === selectionConstants.WORKFLOW_NODE_RESULTS) {
     const outputResultsIndex = ioUtils.getIndexByExtension(
-      props.workflow.run.outputs, '.json',
+      props.workflow.run.outputs, 'results.json',
     );
     let resultValues;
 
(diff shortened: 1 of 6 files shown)


client/public/js/components/status.jsx
```diff
@@ -93,16 +93,18 @@ function Status(props) {
...
     let resultUnit;
+    let resultValue;
 
     if (outputResultsIndex !== -1) {
       const outputResults = props.workflow.run.outputs.get(outputResultsIndex)
         .fetchedValue;
 
       if (outputResults.singlet_energy) {
-        resultValue = outputResults.singlet_energy.value;
+        resultName = outputResults.singlet_energy.name;
```
justinmc: Does this look doable for you to send in the outputs to step1?  Do we want to take this one step further and make there be a dynamic number of output values/names/units to display?

avirshup: Yes and yes!

justinmc:
Should receive something like this for the output values:

   output_values: [
    {
      value: '1234',
      unit: 'hartree',
      name: 'Output Energy'
  ]

justinmc:
Got the 👍 from Aaron via slack.

justinmc:
Ticked off a box in issue https://github.com/Autodesk/molecular-design-applications/issues/76

PR #142, opened 2017-02-14
UX Tweaks for first release

avirshup:


client/public/js/components/workflow_steps.jsx
```diff
@@ -89,21 +91,23 @@ function WorkflowSteps(props) {
           />
           {resultsNode}
         </ol>
+        <Button
+          type="raised"
+          onClick={props.clickRun}
+          disabled={runDisabled}
+          throb={!runDisabled && !finished}
```
justinmc: Excellent job figuring out React :)

avirshup: Heh, we'll see what I ended up screwing up, I'm sure there's something

PR #143, opened 2017-02-14
Email pics

justinmc:
This PR fixes broken images in the emails.

screen shot 2017-02-14 at 9 57 50 am

### DIFF
--- a/client/public/js/components/workflow.jsx
+++ b/client/public/js/components/workflow.jsx
@@ -49,7 +49,8 @@ function Workflow(props) {
     );
   }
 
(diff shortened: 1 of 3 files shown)


avirshup:
@justinmc Looks great! One other related email issue that might be good to go here:

The URL in the emails from Prod and Dev was showing up as `http://molsim.bionano.autodesk.com//[...]` - the two slashes were generating 404s from the server

justinmc:
Ah I'll get that double slash thing here too, good idea.

justinmc:
@avirshup That should fix the double slash thing too.

PR #144, opened 2017-02-14
No local redis cache for prod machines

dionjwa:


dionjwa:
I tested the local dev case, works same as before.

justinmc:
So production will no longer reset the database on every deploy with this?

dionjwa:
I'm merging this to test dev ASAP, time pressure.

justinmc:
👍

PR #145, opened 2017-02-14
Collecting more text edits

avirshup:


avirshup:
@justinmc ready for review here

PR #146, opened 2017-02-14
App bar fixes

justinmc:
A few style fixes to the app bar (see commit names) and got initial load hash scrolling working on the homepage.


justinmc:

PR #147, opened 2017-02-14
Fix Document Page Title

justinmc:
@florencio-mazzoldi noticed that the page title was the undescriptive "Workflow", this changes it to "Molecular Simulation Tools".


justinmc:

PR #148, opened 2017-02-14
'try different tool' button underline

justinmc:
Remove undlerine on 'try different tool' button!


justinmc:

PR #149, opened 2017-02-14
Fix eslint error on line lenght

justinmc:


justinmc:
Quick fix for the deploy, no code review with Aaron's permission.

PR #150, opened 2017-02-14
Status Pane Loading Content

justinmc:
Previously, the status pane still showed the email input stuff after clicking "Run", when the loading spinner was up.  This hides it during that time.

### DIFF
--- a/client/public/js/components/status.jsx
+++ b/client/public/js/components/status.jsx
@@ -15,126 +15,128 @@ require('../../css/status.scss');
 function Status(props) {
   let selection;
 
-  if (props.selection.type === selectionConstants.NODE) {
-    const node = props.nodes.get(props.selection.id);
-    selection = (
-      <div className="status-info">
-        <p>Node</p>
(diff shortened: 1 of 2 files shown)


client/public/js/components/status.jsx
```diff
@@ -15,126 +15,128 @@ require('../../css/status.scss');
...
-    let output;
-    if (lastWorkflowNode && lastWorkflowNode.outputs.size) {
-      output = (
-        <p>
-          Output:
-          <a href={lastWorkflowNode.outputs.get(0).get('value')}>
-            {lastWorkflowNode.outputs.get(0).get('value')}
-          </a>
-        </p>
+  if (!props.fetching) {
```
justinmc: All I did was add this "if" and then indent everything, sorry for the huge diff!

justinmc:
Assuming this is too late for our crunch :)

justinmc:
What I'm talking about is just the little half second loading state here.

loading

PR #152, opened 2017-02-14
Dev/prod URLs

justinmc:
Previously, due to the env vars, the server though it was running on production even when it was on dev.  This PR distinguishes the two.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -6,8 +6,8 @@ services:
     environment:
       CCC: "ccc.bionano.autodesk.com:9000"
       NODE_ENV: "production"
-      FRONTEND_URL: "http://molsim.bionano.autodesk.com"
-      URL: "http://molsim.bionano.autodesk.com"
+      FRONTEND_URL: "http://molsim${BNR_ENV_URL_SUFFIX}.bionano.autodesk.com"
+      URL: "http://molsim${BNR_ENV_URL_SUFFIX}.bionano.autodesk.com"
       REDIS_HOST: "redis.${BNR_ENVIRONMENT}.bionano.bio"
     logging:
       driver: json-file


docker-compose.prod.yml
```diff
@@ -6,8 +6,8 @@ services:
     environment:
       CCC: "ccc.bionano.autodesk.com:9000"
       NODE_ENV: "production"
-      FRONTEND_URL: "http://molsim.bionano.autodesk.com"
-      URL: "http://molsim.bionano.autodesk.com"
+      FRONTEND_URL: "http://molsim${BNR_ENV_URL_SUFFIX}.bionano.autodesk.com"
+      URL: "http://molsim${BNR_ENV_URL_SUFFIX}.bionano.autodesk.com"
       REDIS_HOST: "redis.${BNR_ENVIRONMENT}.bionano.bio"
```
dionjwa: I think you're missing a ".": URL: "http://molsim.${BNR_ENV_URL_SUFFIX}.bionano.autodesk.com"

justinmc: @PeterRJones Does `BNR_ENV_URL_SUFFIX` start with a dot like `.dev`?

PeterRJones: @dionjwa This a different env variable, which includes the `.` so that, when undefined, the environment section of the url vanishes.  When defined it has values of `.dev`, `.qa` and '' (prod)

@justinmc let me merge this. need to fixup something on the automation side to be paired with this change.

justinmc: @PeterRJones sounds good I won't merge it then, feel free to merge it yourself whenever.

PeterRJones: Verified in production that this is fixed/working in template (Cf-Template-asg-dockder-v2.3.8)

PR #153, opened 2017-02-15
Prep.json Validation

justinmc:
It's possible that step0 succeeds in executing, but the python indicates an invalid input by `success: false` in prep.json.  In that case we should not allow the input to be used for the workflow.

screen shot 2017-02-14 at 4 03 26 pm

### DIFF
--- a/client/public/js/actions.js
+++ b/client/public/js/actions.js
@@ -178,6 +178,7 @@ export function selectInputFile(file, workflowId) {
       dispatch({
         type: actionConstants.INPUT_FILE_COMPLETE,
         error: err ? (err.message || err) : null,
+        inputs: err ? err.inputs : null,
       });
     }
   };
@@ -215,6 +216,7 @@ export function submitInputString(input, workflowId) {
       dispatch({
         type: actionConstants.PROCESSED_INPUT_STRING,
         error: err.message || err,
+        inputs: err ? err.inputs : null,
       });
     }
   };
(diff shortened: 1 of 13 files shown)


justinmc:
Oh nice!  I will display that error to the user, should be much more informative.

avirshup:
Cool, thanks. Oh, and actually one more thing - could we still display the 3D structure, even if validation fails? That way the user might get some insight into what's going wrong.

justinmc:
Yeah good call!  I'll take a look at that, shouldn't be too hard.

avirshup:
Sounds good - would be good to take care of that with this PR if it's easy, if not we can add it to the backlog

justinmc:
@avirshup I think I got both of those points now, let me know if this works.

avirshup:
@justinmc Cool! Will test as soon as I've got enough reliable internet! For that last commit, if you haven't already, you'll want to handle the case where `prep.pdb` doesn't exist; some failures may not even have a structure to visualize

justinmc:
In that case it looks like it should show the error from prep.json, not show a structure, and not allow the user to continue.

avirshup:
@justinmc - looks great!  One _small_ request: could we have the error message stay up until the user enters new input? Or, if that would be a new task in itself, maybe just have the message stay up for longer, like 10 seconds.

justinmc:
@avirshup Right now the delay for those snackbar messages is 2 seconds, which I've been feeling is kind of short anyway.  I think I'll bump the delay for snackbar messages across the board in this PR.  I also need to make the input values persist (from the Punchlist issue https://github.com/Autodesk/molecular-design-applications/issues/76), which will make hiding the errors on new input easier, so I'll look into that there.

justinmc:
@avirshup Bumped to 8 seconds after talking to Malte.  Let me know if you think this is good to go.

avirshup:
Great! Good to merge.

PR #155, opened 2017-02-15
Trajectory Animation

justinmc:
This PR will animate the morph feature on the results page!


avirshup:
Note - I updated "Morph" to "trajectory" in the title, since we'll have another set of stories coming in with a different definition of "Morph"

justinmc:
@avirshup It looks like the molviewer can only animate in CMK or stick modes.  Does this look ok to you?  Also it seems like the second frame is all jagged, not sure if that's right.

morphintime

avirshup:
Still looks great! I think @dane1122 knows how to fix the jagged bond problem - that's an issue with the centers of the bonds not getting updated.

dane1122:
@justinmc When I had the jagged bond issue. calling addAnimationFrame(molModelID, stateObj, calcBonds) with the third argument (__calcBonds__) set to _true_ fixed the problem for me!

justinmc:
@dane1122 That fixed it, thank you!

justinmc:
@avirshup I'm merging this after fixing the jagged thing and a few other pieces of cleanup, I hope that's alright.

PR #156, opened 2017-02-15
Logging v2

dionjwa:
- App-wide logging to console and fluent.
 - Crash handler that ensures that crashes are logged to fluent

### DIFF
--- a/server/bin/www
+++ b/server/bin/www
@@ -35,8 +35,6 @@ const streams = [
   },
 ];
 
-console.log('FLUENT_PORT=' + parseInt((process.env['FLUENT_PORT'] || '24226') + ''));
-
 var fluentLogger = require('fluent-logger').createFluentSender('mst', {
   host: 'fluentd',
   port: parseInt((process.env['FLUENT_PORT'] || '24226') + ''),
@@ -115,7 +113,9 @@ Promise.all(seedData.map(workflow =>
   server.on('error', onError);
   server.on('listening', onListening(server.address()));
   server.timeout = 8 * 60 * 1000; // 8 minutes
-}).catch(console.error.bind(console));
+}).catch(err => {
+  log.error({error:err});
+});
 
 /**
  * Normalize a port into a number, string, or false.
--- a/server/bin/www
+++ b/server/bin/www
@@ -228,7 +228,7 @@ process.on('uncaughtException', (err) => {
     errorString: errorString,
     message: 'crash',
   }
-  console.log(JSON.stringify(errObj, null, ''));
+  console.error(JSON.stringify(errObj, null, ''));
   //Ensure crash is logged to fluent before exiting.
   try {
     fluentLogger.emit('mst', errObj, null, () => {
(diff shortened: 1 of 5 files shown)


server/bin/www
```diff
@@ -190,3 +200,41 @@ if (process.env["CCC"] == 'ccc:9000') {
...
+process.on('uncaughtException', (err) => {
+  var stack = null;
+  try {
+    stack = err.stack;
+  } catch(e){};
+
+  //You can't pass the error object directly, fluent doesn't like it.
+  var errorJson = null;
+  try {
+    errorJson = err.toJSON();
```
justinmc: Just checking, this is for things like Dates and not strings, right?

server/bin/www
```diff
@@ -190,3 +200,41 @@ if (process.env["CCC"] == 'ccc:9000') {
...
+    errorString = err.toString();
+  } catch(e){};
+
+  const errObj = {
+    stack: stack,
+    errorJson: errorJson,
+    errorString: errorString,
+    message: 'crash',
+  }
+  console.log(JSON.stringify(errObj, null, ''));
```
justinmc: Would this be better as console.error?  Just to send this to stderr instead of stdout, otherwise I think it doesn't matter.

server/utils/api_utils.js
```diff
@@ -35,7 +35,7 @@ const apiUtils = {
         if (err) {
           return reject(err);
         }
-        console.log(JSON.stringify(response, null, 4));
+        log.debug(JSON.stringify(response, null, 4));
```
justinmc: If you don't stringify this here will it get stringified by your handler code later?

server/utils/run_utils.js
```diff
@@ -112,8 +119,8 @@ const runUtils = {
   },
 
   executeWorkflow(workflowId, email, inputs) {
-    const log = global.log.child({f:'executeWorkflow', workflowId:workflowId, email:email});
-    log.debug({});
+    const localLog = log.child({f:'executeWorkflow', workflowId:workflowId, email:email});
+    localLog.debug({});
```
justinmc: Is this empty log here just saying that the localLog `executeWorkflow` has started?

server/utils/run_utils.js
```diff
@@ -184,26 +191,29 @@ const runUtils = {
...
             }
           }, (err) => {
-            console.log(err);
+            log.error(err);
           });
       });
     }).catch((err) => {
-      console.error(err);
+      log.error(err);
     });
```
justinmc: These can be slightly more compact like `.catch(log.error.bind(log))` FYI, but either way works fine.

server/bin/www
dionjwa: Yes, it's mostly so the fluent daemon can diges objects. Raw Error objects for some reason cause the fluentd msgpack unpacking to barf, and the errors are then lost.

server/bin/www
dionjwa: Good catch, fixed.

server/utils/run_utils.js
dionjwa: Yes

server/utils/api_utils.js
dionjwa: I don't know why I did that. I'm just logging the object now.

PR #158, opened 2017-02-17
Don't Clear Input Strings on Rerender

justinmc:
The strings in the load input and the email input disappear when rerendering or opening the results page.  This will keep them around (at least in some cases, still trying to figure out exactly when we want them and when we don't).

### DIFF
--- a/client/public/css/status_load.scss
+++ b/client/public/css/status_load.scss
@@ -1,5 +1,8 @@
 .status-load {
-  max-width: 100%;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  width: 100%;
   word-wrap: break-word;
 
   .enterMolecule {
--- a/client/public/css/status_results.scss
+++ b/client/public/css/status_results.scss
@@ -2,8 +2,7 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-  width: 215px;
-  margin-top: -60px;
+  width: 100%;
 
   .stat {
     margin-bottom: 32px;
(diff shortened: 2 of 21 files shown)


justinmc:
What this does:
  - Saves submitted input and email strings.  Switching to another step and then returning will still show the submitted value filled in the input.

What this doesn't do:
  - Will NOT save unsubmitted text in either input.  Type in a random string without submitting, switch to another step and return, and the string will be reset to either empty string or the last submitted value.
  - Will NOT populate the results page with values entered pre-run.  On the results page, the inputs will be empty.

I'm concerned about the results page and how it relates to the setup page.  @avirshup I think you and I should talk through how they relate and all the edge cases.  I could do fancier stuff with saving the input values in this PR, but it starts to be weird behavior or possibly irrelevant work on the results page.

input_preservation
email_preservation

justinmc:
After talking with @avirshup, in order to solve the problems that things like this are building up on the results page, I will disable the ability to edit and run on the results page for now.  That's coming soon in this PR.

justinmc:
Todo:

  - [ ] Populate Load input on results page
  - [ ] Show a download button at the bottom of Load

justinmc:
@avirshup I want to check with you one more time before I merge.  I've disabled the ability to run on the results page, populated the inputs on the results page, and displayed a download button on the "load" step.  The new download button looks like this:

screen shot 2017-02-22 at 12 42 32 pm

I also changed the results step download button to say "Download Result" to clarify that the two download buttons aren't downloading the same thing.

PR #175, opened 2017-02-22
Cancel UI

justinmc:
This PR implements Malte's design for a cancel confirmation dialog and cancel page.  It also fixes a bug that had broken cancels.

screen shot 2017-02-21 at 4 33 26 pm

screen shot 2017-02-21 at 4 30 30 pm

### DIFF
--- a/server/utils/run_utils.js
+++ b/server/utils/run_utils.js
@@ -66,6 +66,13 @@ const runUtils = {
     // Set the final output and status on the run
     return redis.hget(dbConstants.REDIS_RUNS, runId).then((runString) => {
       const run = JSON.parse(runString);
+
+      // Don't set results on a canceled run
+      if (run.status === statusConstants.CANCELED) {
+        localLog.debug(`Run ${runId} canceled, so results not written.`);
+        return Promise.resolve();
+      }
+
       const status = jobResult.exitCode === 0 ?
         statusConstants.COMPLETED : statusConstants.ERROR;
       var outputs = [];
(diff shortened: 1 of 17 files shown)


server/utils/run_utils.js
```diff
@@ -66,6 +66,13 @@ const runUtils = {
     // Set the final output and status on the run
     return redis.hget(dbConstants.REDIS_RUNS, runId).then((runString) => {
       const run = JSON.parse(runString);
+
+      // Don't set results on a canceled run
+      if (run.status === statusConstants.CANCELED) {
+        localLog.debug(`Run ${runId} canceled, so results not written.`);
+        return Promise.resolve();
+      }
+
```
justinmc: @dionjwa Can you sign off on this?

I noticed that canceling a workflow wasn't quite working.  It looks like our "cancel" endpoint is not actually canceling runs anymore, it just marks them as canceled in the database.  The run still continues, and when it finishes, it overwrites the cancel with "success".

As a temporary fix, this code here skips writing anything to a canceled run.  The "canceled" run will still finish running, but nothing will be written to the db, so the user will always still see the canceled page at the run's url.  I think the permanent solution will come once we've moved status out of the database and just fetch it from CCC like we discussed, but we'll still need some sort of CCC endpoint to be able to cancel the job for real.

dionjwa: There is a cancel CCC endpoint, but I think it would be better to wait until the refactoring in https://zube.io/autodesk/molecular-design-applications/c/163 before actually do it.

justinmc: Cool I agree with that.  I'll go with this hacky solution for now and will make a note to make sure cancel works when we refactor to get the status from CCC

PR #178, opened 2017-02-22
Results page show results step by default

justinmc:
When loading the results of a run, the results step should be selected by default.


avirshup:
Despite my trouble with the GitHub reviews, this looks great! :+1: to merge

PR #180, opened 2017-02-24
Molviewer Toolbar

justinmc:
I was unable to include the main molviewer CSS because it would shrink the viewer in weird ways.  Andrew kindly made me a tiny css file for the headless viewer only that allows us to show the missing toolbar icons.

In the future, this css will ideally be a part of the molviewer package instead of its own css file in this repo.

screen shot 2017-02-24 at 3 06 36 pm

### DIFF
--- a/client/public/css/molviewer.css
+++ b/client/public/css/molviewer.css
@@ -1,0 +1,218 @@
+.adsk-toolbar {
+    display: block;
+    position: absolute;
+    left: auto;
+    right: 0;
+    bottom: auto;
+    top: calc(50% - 40px);
+    text-align: center;
+    transition: opacity 0.2s ease 0.5s;
+    box-sizing: content-box;
+    z-index: 2;
+    pointer-events: none;
+}
+
+.notouch .adsk-toolbar {
+    opacity: 0.9;
+}
+
+.adsk-control-group {
+    display: block;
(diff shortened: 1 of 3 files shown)


justinmc:
@avirshup Whoops forgot to add that file to the commit :)

avirshup:
@justinmc I'm actually getting the bug from #179 here, so I'm not able to see the controls :(

justinmc:
@avirshup Ah true, this will have to wait until we've got the molviewer thing fixed.

justinmc:
@avirshup The molviewer bug should be gone now, so this should be reviewable.

avirshup:
@justinmc - ok, got this up and running with tirrenu-1.3.5! The toolbar's working well. However, the protein displays in stick mode now:
image
But I'll try some newer branches and see if the problem manifests there too

justinmc:
@avirshup I put everything in stick mode in https://github.com/Autodesk/molecular-simulation-tools/pull/155 due to the animation API needing to be in stick mode.  Maybe I can do something fancier and only show stick mode on the results step if that's better.

avirshup:
Ohhh, right, I remember that. In which case - this looks good!

justinmc:
Ok cool.  We can always do another PR to sort out the stick stuff.

Issue #204, opened 2017-03-06
UI should be generated dynamically by app format


justinmc:
Instead of hardcoding the UI parts of the workflow, the frontend should receive some data that describes the UI to be displayed.  Individual steps will be hardcoded in and referenced by the app data (i.e. ligand selection).  Only currently existing UI is necessary to be implemented for this card, but future UI pieces should be considered.

justinmc:
PR: https://github.com/Autodesk/molecular-simulation-tools/pull/231

Though note there will be lots more to do to take this forward.  A lot is hardcoded in that PR, as noted in this issue's description.

avirshup:
Keeping this open in the new sprint to track work and PRs on this issue

justinmc:
Two more PRs working on this:

https://github.com/Autodesk/molecular-simulation-tools/pull/263
https://github.com/Autodesk/molecular-simulation-tools/pull/265/

PR #228, opened 2017-03-13
Generalize selectedLigand

justinmc:
This PR is a refactor that sets us up to move to a generic, dynamic step UI.  `selectedLigand` previously was duplicated state, stored in an inconvenient place in selection.json as given by the server, and also stored directly on the run.  This gets rid of the direct storage on the run so there is only one source of truth, and we can rely more on using server inputs for generic storage in the future.

Part of issue https://github.com/Autodesk/molecular-simulation-tools/issues/204

### DIFF
--- a/client/public/js/actions.js
+++ b/client/public/js/actions.js
@@ -69,14 +69,16 @@ export function initializeRun(workflowId, runId) {
       outputs = await workflowUtils.fetchIoPdbs(outputs);
       outputs = await workflowUtils.fetchIoResults(outputs);
 
-      // Find the selected ligand in the inputs, if it exists
-      const selectedLigand = ioUtils.getSelectedLigand(inputs);
+      // If only one ligand, select it
+      const ligands = ioUtils.getLigandNames(inputs);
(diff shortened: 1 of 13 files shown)


dionjwa:
Hey Justin, what is the goal with this? I was thinking that we would have the client simply call a completely generic API on the server that simply performs a fast docker compute and returns the results. In other words, the server would have no idea about 'ligands', since that is highly specific to a specific workflow.

justinmc:
@dionjwa This PR is just a small tweak to frontend-only stuff, so we should be on track for that plan.  The frontend receives a bunch of inputs and outputs json and stores it all.  It used to pull out the specific selected ligand string and store that separately as well, but now with this PR it just accesses it within the inputs/outputs json directly.

I think we're moving in the right direction toward a generic solution where the server doesn't know about specific tasks.  Not sure how @avirshup generates selection.json and prep.json containing the list of ligands, but that's outside of the Node server at least.

dionjwa:
Got it.

PR #230, opened 2017-03-14
Fix MolViewer Bugs on Destruction

justinmc:
I was seeing lots of errors popping up in corner cases for the molviewer: switching morph frames quickly, and navigating onto and off of pages with the molviewer.  This PR cleans up the sketchiness of that code by wrapping the asynchronous complexity of the molviewer into a wrapper class.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -18,7 +18,7 @@
     "react-tap-event-plugin": "^2.0.1",
     "redux": "^3.6.0",
     "redux-thunk": "^2.1.0",
-    "tirrenu": "../molecule_viewer/tirrenu-1.3.2.tgz",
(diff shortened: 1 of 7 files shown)


justinmc:
@avirshup I just updated the molviewer to the latest version, and after talking with @ktbolt, it sounds like the selection strings that you're giving me are no longer supported. So currently the first ligand in 3ac2 is given to me like this:

```
mv_ligand_strings: {
  "SO4 (seq # 510)": ["1.A.A-510"],
  ...
}
```

But that string `1.A.A-510` will no longer work as is.

justinmc:
@avirshup The selection string bug will start cropping up in master now that this is merged, but I think we'll have to fix that elsewhere.

PR #229, opened 2017-03-14
Run step

justinmc:
This PR combines the email step and the run button into a run step.

Issue: https://github.com/Autodesk/molecular-simulation-tools/issues/184
Also part of doing dynamic steps: https://github.com/Autodesk/molecular-simulation-tools/issues/204

screen shot 2017-03-14 at 11 12 04 am
screen shot 2017-03-14 at 11 12 18 am
screen shot 2017-03-14 at 11 11 22 am

Issue #205, opened 2017-03-06
Define App Data


justinmc:
We should define an app data format that will be used for describing the UI that the frontend will generate (see https://github.com/Autodesk/molecular-design-applications/issues/204).

justinmc:
Thinking out loud, here's an MVP way we could do this:

  * Workflows are the top-level data structure as they are now.
  * Workflows have at least one WorkflowNode.  Each one corresponds to a possible "step" that you see numbered on the left side of the app right now.
    - WorkflowNodes are hardcoded into the app.
    - Each has a unique string id (basically a name)
    - Workflows have a `workflowNodeIds` field that is an array of strings.
  * The UI is generated based on these workflowNodes.
  * Data from the user and from the server that corresponds to each workflowNode is set on the `inputs` array, as it is today.
  * The final submit step is also a workflowNode itself?

justinmc:
One step above MVP: If we need to do any configuration on workflowNodes, maybe the `workflowNodeIds` array could instead of just ids be objects with an `id` parameter and some configuration parameters.

joseph-schaeffer:
I like this setup, but "Workflows" is a pretty overloaded term, and the way I see you using them here are as application steps rather than specific computational workflows. If we could come up with another term (much like front-end app steps are now "Widgets") to keep it from colliding with the computational workflow, that would be great.

joseph-schaeffer:
Hit submit too soon. Other than the comment above, I think the steps you've described make sense for the app setup and definition, with one possible exception - I heard at least one requirement for a step that doesn't show up, but does invoke some computation (front-end widget, or back-end workflow). I believe that's being used for input validation/manipulation in some of the current apps.

justinmc:
I agree!  These "workflowNodes" might not even have any backend computation in some cases.  Hopefully we can nail down the nomenclature a bit tomorrow.

I'll bring up the "invisible" steps thing tomorrow as well and make sure that our solution can do that.  Thanks for taking the time to look at this.

dionjwa:
I agree with Joseph in that we need to agree on the terminology. Front end 'widgets' or  'components' or whatever you want to call them should not be called workflows. They do not obey some important assumptions implied with the term 'workflow'.

dionjwa:
[...]

PR #231, opened 2017-03-14
Dynamic Steps

justinmc:
This PR will be my first stab at generating the "tasks" steps seen numbered on the frontend dynamically, based on data passed from the server with the workflow.  I'll try to limit the backend impact of this for now.

Issues https://github.com/Autodesk/molecular-simulation-tools/issues/204 and https://github.com/Autodesk/molecular-simulation-tools/issues/205

~~Waiting on https://github.com/Autodesk/molecular-simulation-tools/pull/228 and https://github.com/Autodesk/molecular-simulation-tools/pull/229~~


justinmc:
@PeterRJones Check out the migration setup I wrote for this: https://github.com/Autodesk/molecular-simulation-tools/pull/231/commits/31327b8fb4667564b2cb78e10f918ac1256ec02f#diff-7cd651498ecb2e6d55bc2ea9aa2785e0R30

How do you normally handle db migrations?  Is the production db nicely backed up in a way that we can roll back if these migrations destroy everything?  It worked on my local db but I'm nervous :)

client/public/js/components/tasks.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <div className="tasks-pane">
+      <div key={0} className="tasks">
+        <ol>
+          {
+            props.app.tasks.map((task, index) => {
+              const number = index + 1;
+
+              switch (task.id) {
```
justinmc: The tasks array is iterated, and task ids are converted into an actual UI button (left column) in a big switch statement.  Might be moved into a utils file later, but the general idea is the same.

screen shot 2017-03-22 at 10 17 52 am

client/public/js/utils/task_utils.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+  },
+
+  /**
+   * Given a task (and run data), return a bool indicating if it's completed
+   * @param {TaskRecord}
+   * @param {RunRecord}
+   * @return {Boolean}
+   */
+  isCompleted(task, run) {
```
justinmc: Logic for determining the completion state of a task is here.  There is no separate state for marking that a task is completed.  It's just determined by looking at run state (i.e. for Load, if input pdb data exists, then Load must be done).

client/public/js/components/status.jsx
```diff
@@ -115,32 +35,70 @@ function Status(props) {
...
-        <StatusLigandSelection
-          changeLigandSelection={props.changeLigandSelection}
-          ligandNames={ioUtils.getLigandNames(props.workflow.run.inputs)}
-          runCompleted={runCompleted}
-          selectedLigand={props.selectedLigand}
-        />
-      );
+    } else if (!props.app.fetching && !props.app.fetchingError &&
+      props.selection.type === selectionConstants.TASK) {
+      switch (props.app.tasks.get(props.selection.taskIndex).id) {
```
justinmc: The status pane finds its content the same way as the numbered steps in the left column.  A switch statement on the task id runs some custom code for each task id.

screen shot 2017-03-22 at 10 19 10 am

server/utils/seed_data.js
```diff
@@ -10,6 +12,14 @@ const workflows = [
     creatorImage: '/img/logo2.png',
     description: 'Calculate the electron binding energy of an' +
         ' anionic doublet species using DFT',
+    tasks: [
```
justinmc: The only task data we're using for now is the task ID, so that's all I'm including in the app data.

justinmc:
@dionjwa Please take a look at this as a first step on the way towards using the dynamic app data that we've been discussing.  I tried to add comments on the interesting parts since the PR is huge (due to renaming `workflow` to `app` everywhere).  I left the name `tasks` as is since we didn't seem conclusive on a final name yesterday.  Also notice that I'm purposely not yet using an intermediate VERIFY step for loading a PDB.

client/public/js/components/app_card.jsx
```diff
@@ -72,17 +74,17 @@ function WorkflowCard(props) {
...
 
-WorkflowCard.propTypes = {
+AppCard.propTypes = {
   bgColor: React.PropTypes.string.isRequired,
   bgIndex: React.PropTypes.number.isRequired,
   color: React.PropTypes.string.isRequired,
   comingSoon: React.PropTypes.bool.isRequired,
-  creatorImage: React.PropTypes.string.isRequired,
+  // creatorImage: React.PropTypes.string.isRequired,
   id: React.PropTypes.string.isRequired,
```
dionjwa: Just remove?

dionjwa:
@justinmc This is too big with too many changes, about changes that are probably only partially done, for me to review this effectively. I'm going to merge, but you'll need to guide me through these current changes and the proposed changes.

PR #232, opened 2017-03-15
Server Eslint

justinmc:
This PR gets the server code to pass eslint, and it also requires eslint to pass in order to run `npm run dev`.

### DIFF
--- a/server/bin/www
+++ b/server/bin/www
@@ -62,6 +62,7 @@ function onError(error) {
   // handle specific listen errors with friendly messages
   switch (error.code) {
     case 'EACCES':
+      // This should never happen when running in docker
       log.error(`${bind} requires elevated privileges`);
       process.exit(1);
       break;
--- a/server/bin/www
+++ b/server/bin/www
@@ -4,7 +4,6 @@
  * Module dependencies.
  */
 require('dotenv').config();
-require('debug')('express:server');
 const http = require('http');
 const httpProxy = require('http-proxy');
 const retry = require('bluebird-retry');
--- a/server/bin/www
+++ b/server/bin/www
@@ -86,7 +86,7 @@ function onListening(addr) {
     log.debug(`Listening on ${bind}`);
     // Wait a bit, sometimes docker takes a while
     retry(
-      () => testUtils.runTestCCC(),
+      testUtils.runTestCCC,
       { max_tries: 20, interval: 1000 }
     ).then((status) => {
       log.info({ message: 'Test CCC connection', success: status.success });
--- a/server/bin/www
+++ b/server/bin/www
@@ -4,7 +4,6 @@
  * Module dependencies.
  */
 require('dotenv').config();
-require('babel-polyfill');
 require('debug')('express:server');
 const http = require('http');
 const httpProxy = require('http-proxy');
--- a/server/bin/www
+++ b/server/bin/www
@@ -3,96 +3,95 @@
 /**
  * Module dependencies.
  */
-const dbConstants = require('../constants/db_constants');
-const dbUtils = require('../utils/db_utils');
-const seedData = require('../utils/seed_data');
-const redis = require('../utils/redis');
-
-if (!('toJSON' in Error.prototype))
-  Object.defineProperty(Error.prototype, 'toJSON', {
-    value: function () {
-    var alt = {};
-
-    Object.getOwnPropertyNames(this).forEach(function (key) {
-      alt[key] = this[key];
-    }, this);
-
-    return alt;
-  },
-  configurable: true,
-  writable: true
-});
-
 require('dotenv').config();
 require('babel-polyfill');
-const bunyan = require('bunyan');
-const bunyanFormat = require('bunyan-format');
-
-const streams = [
-  {
-    level: process.env['NODE_ENV'] == "development" ? bunyan.TRACE : bunyan.DEBUG,
-    stream: bunyanFormat({ outputMode: 'short' }),
-  },
-];
-
-var fluentLogger = require('fluent-logger').createFluentSender('mst', {
-  host: 'fluentd',
-  port: parseInt((process.env['FLUENT_PORT'] || '24226') + ''),
-  timeout: 3.0,
-  reconnectInterval: 600000, // 10 minutes
-});
-
-streams.push({
-  level: bunyan.DEBUG,
-  type: 'raw',// use 'raw' to get raw log record objects
-  stream: {write:(obj, cb) => {
-    switch(typeof(obj)) {
-      case 'object':
-        break;
-      default:
-        obj = {message:toString(obj)};
-    }
-    if (obj['time']) {
-      obj['@timestamp'] = obj['time'].toISOString();
-    }
-    if (!obj['@timestamp']) {
-      obj['@timestamp'] = new Date().toISOString();
-    }
-    delete obj['time'];
-    if (obj['error'] && typeof(obj['error']) == 'object') {
-      try {
-        obj['error'] = JSON.stringify(obj['error']);
-      } catch(e){}
-    }
-    if (obj['err'] && typeof(obj['err']) == 'object') {
-      try {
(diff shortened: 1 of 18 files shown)


justinmc:
@dionjwa Check this out, I mainly just cleaned up a million little spaces and commas and stuff, but I did move a few things to separate files.  Also just to check, I'm ok to add eslint to the npm dev script here, right?  It's only run in development?

dionjwa:
Why don't we create a git pre-commit hook? It's pretty straightforward to create them, and link them locally, and to me it makes a bit more sense to have the linting check on the commit step rather than the dev server, since you can run the server, make a lint-breaking change, them commit.

justinmc:
I think I'll move the lint/nodemon combo into another script that I can use if I want, since the frontend webpack-dev-server already runs eslint on each code change too.  I was playing just playing around with pre-commit hooks and this seemed to work for me in `.git/hooks/pre-commit`:

```
cd ./server/
exec npm run lint
```

justinmc:
@dionjwa Done.

server/bin/www
```diff
@@ -3,126 +3,27 @@
...
-    }, this);
-
-    return alt;
-  },
-  configurable: true,
-  writable: true
-});
-
 require('dotenv').config();
 require('babel-polyfill');
```
justinmc: I think I added it just to make sure we had the same javascript features available on the server and the client.  I couldn't find any specific code that needed it or any other reason from looking at the PR where I added it.  I also tried removing it and everything seems to still work, so I'll push a commit that removes it.

server/bin/www
```diff
@@ -3,126 +3,27 @@
...
-
-global.log = bunyan.createLogger({
-  name: 'mst',
-  level: process.env['NODE_ENV'] == "development" ? bunyan.TRACE : bunyan.DEBUG,
-  streams,
-});
-
-const app = require('../main/server');
-const debug = require('debug')('express:server');
+require('debug')('express:server');
```
justinmc: It looks like they both do similar things, but this one is no longer used since we added bunyan.  Removing!

server/bin/www
```diff
@@ -3,126 +3,27 @@
...
+const retry = require('bluebird-retry');
+const app = require('../main/server');
+const dbConstants = require('../constants/db_constants');
+const dbUtils = require('../utils/db_utils');
+const errorUtils = require('../utils/error_utils');
+const log = require('../utils/log');
+const fluentSender = require('../utils/fluent_sender');
+const redis = require('../utils/redis');
+const seedData = require('../utils/seed_data');
+const testUtils = require('../test/test_utils');
```
justinmc: Good point.  I'm so used to this that I don't even notice anymore haha.  Are you thinking absolute paths from the project root would be better, or single unique names?  I'm unaware of any nice ways to do either one.  I think all node apps I've seen use relative paths like this, and even the new frontend import/export spec works with relative paths.

I guess one thing this app does do fairly well is keep everything in a folder 1 level deep, so pretty much every require will include just a single `../` and will work in any other folder.  I can keep that in mind as I keep developing to make sure I strictly adhere to it.

Could be interesting to look into all the possibilities for improving this though.

server/bin/www
```diff
@@ -138,26 +39,32 @@ function normalizePort(val) {
 }
 
 /**
- * Event listener for HTTP server "error" event.
+ * Get port from environment and store in Express.
  */
+const port = normalizePort(process.env.PORT || '4000');
```
justinmc: True, I hate declaring loose functions in the middle of a file like this too.  This code comes from express's boilerplate generator.  Pushing a commit to inline it.

server/bin/www
```diff
@@ -138,26 +39,32 @@ function normalizePort(val) {
...
+  const bind = typeof port === 'string'
+    ? `Pipe ${port}`
+    : `Port ${port}`;
 
   // handle specific listen errors with friendly messages
   switch (error.code) {
     case 'EACCES':
-      log.error(bind + ' requires elevated privileges');
+      log.error(`${bind} requires elevated privileges`);
       process.exit(1);
```
justinmc: This is more express generator code, but I'm in favor of keeping it in case this is run outside of a docker container.  I've run it outside of docker myself a few times and I'd expect to see this typical express error message if needed.  I'll add a comment at least though.

server/bin/www
```diff
@@ -166,75 +73,95 @@ function onError(error) {
...
-      )
-      .then(status => {
-          log.info({message:'Test CCC connection', success:status.success});
-      })
-      .catch(err => {
-        log.error({status:'Test failed:', error:JSON.stringify(err)});
-      });
-  }
+      () => testUtils.runTestCCC(),
+      { max_tries: 20, interval: 1000 }
```
justinmc: Yes, good catch!

server/bin/www
```diff
@@ -166,75 +73,95 @@ function onError(error) {
...
+  log.error({ error: err });
+});
+
+/**
  * Proxy requests to port 9000 to the CCC server
  */
-if (process.env["CCC"] == 'ccc:9000') {
-  const httpProxy = require('http-proxy');
-  httpProxy.createProxyServer({target:'http://ccc:9000'}).listen(9000);
+if (process.env.CCC === 'ccc:9000') {
```
justinmc: Yeah just a style thing.  Here is the explanation from eslint: http://eslint.org/docs/rules/dot-notation

server/main/version.js
```diff
@@ -1,25 +1,26 @@
...
+// Path to package.json and VERSION files are same - at root of this node application.
 // Use the same logic to find both files.
-// In a typcial container, this usually means '/app/server' 
-const packageJSON = JSON.parse(fs.readFileSync('package.json', {encoding:'utf8'}));
+// In a typcial container, this usually means '/app/server'
+const packageJSON = JSON.parse(fs.readFileSync('package.json', { encoding: 'utf8' }));
 const versionFilePath = 'VERSION';
 
-var newLine = /\n/g;
+const newLine = /\n/g;
```
justinmc: I blame Peter, but yeah that seems more robust, will fix.

server/utils/fluent_sender.js
```diff
@@ -0,0 +1,10 @@
+const fluentLogger = require('fluent-logger');
```
justinmc: It's used in TWO places, so it's totally worth it haha.  Plus this will be available for us to use to log stuff elsewhere.

server/utils/run_utils.js
```diff
@@ -1,13 +1,11 @@
-const fs = require('fs-extended');
-const path = require('path');
 const Promise = require('bluebird');
+const statusConstants = require('molecular-design-applications-shared').statusConstants;
+const cccUtils = require('../utils/ccc_utils.js');
```
justinmc: The trend is to only make newable classes upper camel case, and I think eslint will complain otherwise.

server/utils/workflow_utils.js
```diff
@@ -22,50 +24,49 @@ const workflowUtils = {
...
-    log.debug({execute:'executeWorkflow0Step0', job:JSON.stringify(jobJson).substr(0, 100)});
+    log.debug({ execute: 'executeWorkflow0Step0', job: JSON.stringify(jobJson).substr(0, 100) });
     return workflowUtils.executeCCCJob(jobJson)
-      .then(jobResult => {
+      .then((jobResult) => {
         log.debug(jobResult);
-        var outputs = [];
-        for (var i = 0; i < jobResult.outputs.length; i++) {
+        const outputs = [];
+        for (let i = 0; i < jobResult.outputs.length; i += 1) {
```
justinmc: This was probably the weirdest eslint rule for me too.  ++ has sketchy behavior in javascript in some cases: http://eslint.org/docs/rules/no-plusplus

server/bin/www
dionjwa: I disagree with that one, and the reasoning there is not convincing, but I'll concur to the hive mind

dionjwa:
Looks good!

Issue #259, opened 2017-03-28
Display ribbons view for all steps except results animation


avirshup:
It would be great to have this for prod on Thursday if possible. I'd like to see the default view (ribbons for the protein, sticks for the small molecules) everywhere except the final animation, where all sticks is OK for now.

justinmc:
I'll prioritize this and shoot for Thursday!

justinmc:
As of https://github.com/Autodesk/molecular-simulation-tools/issues/263, this will show "ribbon" before you press Run and "stick" afterwards.  I think this card requires the view to change when you click a given widget though, so it's not done yet.

justinmc:
https://github.com/Autodesk/molecular-simulation-tools/pull/265 will do what I think is the full solution to this card:

![ribbon_slash_stick](https://cloud.githubusercontent.com/assets/389558/24682601/58e6a5c4-194f-11e7-8ca3-407e7ff8057e.gif)

justinmc:
Closing since https://github.com/Autodesk/molecular-simulation-tools/pull/265 has been merged.

PR #263, opened 2017-03-29
More Dynamic Widgets

justinmc:
~~Do not merge/review, waiting on https://github.com/Autodesk/molecular-simulation-tools/pull/231~~

This moves us more in the direction of dynamic widgets:

  - Widgets now have meta data (and get their title from the meta data).
  - "Results" is now also a widget itself, not hardcoded.
  - The widget list on the left side is totally generated dynamically.
  - View is more dynamic and less dependent on hardcoded widgets.

This PR also fixes half of https://github.com/Autodesk/molecular-simulation-tools/issues/259, since pre-results "ribbon" is shown, and "stick" is shown after running, but I still need to change the representation based on the selected widget.


client/public/js/components/status.jsx
```diff
@@ -84,13 +54,48 @@ function Status(props) {
...
+          let resultValues;
+
+          if (outputResultsIndex !== -1) {
+            const outputResults = props.app.run.outputs.get(outputResultsIndex)
+              .fetchedValue;
+
+            if (outputResults.output_values) {
+              resultValues = new IList(outputResults.output_values);
+            }
+          }
```
justinmc: This "status pane" still is controlled by a switch statement on the widgetId, but now Results is its own widget too.

screen shot 2017-03-29 at 11 27 52 am

client/public/js/components/view.jsx
```diff
@@ -1,54 +1,118 @@
...
+        pdbs,
+        nextProps.loading,
+      );
+    }
+    if (!is(selectionStrings.toSet(), oldSelectionStrings.toSet())) {
+      this.renderMoleculeViewerSelection(selectionStrings, pdbs.size);
+    }
+    if (nextProps.morph !== this.props.morph) {
+      this.renderMoleculeViewerPdbIndex(nextProps.morph);
+    }
```
justinmc: view.jsx, which contains the molviewer, now receives all inputs and outputs and decides what to display itself based on those.  This is an intermediate step until view.jsx is controlled more directly by which widget is selected and what that widgets specific inputs/outputs are.

server/utils/seed_data.js
```diff
@@ -32,15 +44,30 @@ const apps = [
...
+        meta: {
+          title: 'Run',
+        },
       },
       {
-        id: tasksConstants.RUN,
+        id: widgetsConstants.RESULTS,
+        meta: {
+          title: 'Results',
+        },
```
justinmc: We're now using these titles in the meta data.

client/public/js/components/widget_list.jsx
```diff
@@ -0,0 +1,56 @@
...
+                key={widget.id}
+                number={index + 1}
+                onClick={props.clickWidget}
+                primaryText={widget.meta.title}
+                selected={props.selection.widgetIndex === index}
+                status={widgetStatuses.get(index)}
+                index={index}
+                totalNumber={props.app.widgets.size}
+              />
+            ))
```
justinmc: Instead of a switch statement, widget list buttons on the far left are just rendered based on their `meta` data title.

screen shot 2017-03-29 at 11 31 29 am

dionjwa:
Looks good. Will be clearer once it's more complete.

PR #264, opened 2017-03-31
Misc cleanup, fix error handling, and update CCC client

dionjwa:


server/bin/www
```diff
@@ -1,6 +1,24 @@
 #!/usr/bin/env node
 
 /**
+ * Add this extra property to Errors right away
+ */
+if (!('toJSON' in Error.prototype))
```
justinmc: Don't forget to add that eslint precommit hook!  There are a few violations here.  You also might have to modify the linter rules to be able to modify Error.prototype, but maybe it's not smart enough to know since you're using `defineProperty`

justinmc:
👍  when it's passing eslint.

server/bin/www
dionjwa: Fixed linting errors!

PR #265, opened 2017-03-31
Widget I/O Data

justinmc:
~~Do not merge, WIP and waiting on https://github.com/Autodesk/molecular-simulation-tools/pull/263.~~

This PR adds inputs and outputs to app data and uses it to make widgets more generic.  For example, the app now determines whether a widget is completed by looking at if it has created all of its output data.

The way that this PR stores "result" data from a widget is in the map at run.ioResults keyed on widget input/output ids.  So the app looks at a widget's inputs or outputs and then looks up the actual result values in run.ioResults.


justinmc:
This needs to be tested on the Results page.  I can't now because runs are not completing.

server/utils/seed_data.js
```diff
@@ -50,27 +71,57 @@ const seedData = {
...
+            { id: 'workflow_state.dill' },
+            { id: 'selection.json' },
+          ],
+          outputs: [
+            { id: 'final_structure.pdb' },
+            { id: 'results.json' },
+            { id: 'minstep.0.pdb' },
+            { id: 'minstep.1.pdb' },
+            { id: 'minsteps.tar.gz' },
+            { id: 'minstep_frames.json' },
```
justinmc: Here I've added inputs/outputs to the app json, which is the main point of this PR.  I named them the same as the outputs that are already being generated by the server.  There is currently no real validation of this data, but if you set it to something that doesn't make sense, it will break.

client/public/js/records/io_record.js
```diff
@@ -1,11 +1,7 @@
 import { Record } from 'immutable';
 
 const IoRecord = new Record({
-  name: '',
-  type: '',
-  value: '',
-  // In case of a url value, the frontend will fetch it and store the result here
-  fetchedValue: '',
+  id: '',
 });
```
justinmc: IoRecords are now bare bones, they're exactly what you see in the app json (id only).  Every widget has a list of inputs and a list of outputs, each input/output represented by an IoRecord.

client/public/js/records/io_result_record.js
```diff
@@ -0,0 +1,11 @@
+import { Record } from 'immutable';
+
+const IoResultRecord = new Record({
+  ioId: '',
+  type: '',
+  value: '',
+  // In case of a url value, the frontend will fetch it and store the result here
+  fetchedValue: '',
+});
```
justinmc: IoResults are now where the meaty run data is stored.  The run contains a map of these IoResults, keyed on ioId.  These ioIds are referenced in IoRecords as a widget defines its inputs and outputs.  The actual data is stored in one place, in this ioResults map, and can be looked up as needed.

client/public/js/utils/widget_utils.js
```diff
@@ -1,68 +1,57 @@
...
-
-      case widgetsConstants.RUN:
-        return run.status === statusConstants.COMPLETED;
-
-      case widgetsConstants.RESULTS:
-        return run.status === statusConstants.COMPLETED;
-
-      default:
-        throw new Error(`Invalid widgetId: ${widget.id}`);
-    }
```
justinmc: No more hardcoded logic like this is needed to know how to check if a widget is completed/active/disabled.  Now we can just check to see if a widget has all of its inputs or outputs available (see `getStatus` above).

client/public/js/actions.js
```diff
@@ -110,18 +113,16 @@ export function clickWidget(widgetIndex) {
...
   return (dispatch) => {
     dispatch({
       type: actionConstants.CLICK_RUN,
     });
 
-    const selectedLigand = ioUtils.getSelectedLigand(inputs);
-
-    apiUtils.run(appId, email, inputs, selectedLigand, inputString).then((runId) => {
+    apiUtils.run(appId, email, inputResults, inputString).then((runId) => {
       dispatch({
```
dionjwa: Can we always indent promise statements? So then `.then(` statements are always on a newline, so that catch statements and other thens align, I find it easier to see the flow control and to distinguish between promises and callbacks.

client/public/js/actions.js
```diff
@@ -158,19 +159,19 @@ export function selectInputFile(file, appId) {
 
     try {
       const inputString = await appUtils.readFile(file);
-      let inputs = await appUtils.processInput(
+      let inputResults = await appUtils.processInput(
         appId, inputString, extension,
```
dionjwa: Is it necessary to disambiguate between inputs and inputResults?

client/public/js/records/io_result_record.js
dionjwa: Is the ioId the "pipeId"? I think the name ioId is a bit ambiguous.

client/public/js/utils/widget_utils.js
dionjwa: Great!

client/public/js/actions.js
justinmc: Yes, will do!  I promised (no pun intended) you I'd do that but I keep forgetting.

client/public/js/records/io_result_record.js
justinmc: Just discussed in person, will rename ios to pipes, and ioId will become pipeId.  Also, ioResult will become pipeData.

client/public/js/actions.js
justinmc: Just discussed in person about renaming ios/ioResults to pipes/pipeDatas.

justinmc:
@dionjwa Just fixed your issues (renamed ios to pipes, ioResults to pipeDatas, and put a return after `.then`).  Will plan to merge this after lunch.

PR #266, opened 2017-03-31
Add explicit git hooks for the server

dionjwa:


justinmc:
I didn't know you could do this, good idea!  👍

PR #267, opened 2017-04-03
Travis working. Will build client and server, create the docker image…

dionjwa:
- Travis working.
     - Will build client and server
     - create the docker image
     - run server functional tests, failing if any fail
     - then push images to quay.io if credentials present and master branch

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,8 +3,8 @@
 .DS_Store
 .jenkins-config
 .env*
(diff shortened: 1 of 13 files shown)


justinmc:
Looks good to me after https://github.com/Autodesk/molecular-simulation-tools/pull/268 goes in 👍 

Let me know if things look good after this is merged with my migration fixes!

PR #268, opened 2017-04-03
Fix Migration Errors

justinmc:
This PR fixes errors that would occur during db initialization for a fresh database.  It also fixes an error logging bug in the process!

### DIFF
--- a/server/bin/www
+++ b/server/bin/www
@@ -5,7 +5,7 @@
  */
 if (!('toJSON' in Error.prototype)) {
   Object.defineProperty(Error.prototype, 'toJSON', {
-    value: () => {
+    value: function toJSON() {
       const alt = {};
 
       Object.getOwnPropertyNames(this).forEach((key) => {
@@ -110,7 +110,7 @@ dbUtils.initialize(redis).then(() => {
   server.on('listening', onListening(server.address()));
   server.timeout = 8 * 60 * 1000; // 8 minutes
 }).catch((err) => {
-  log.error({ error: err });
+  log.error({ error: err.toJSON() });
 });
 
 /**
--- a/server/utils/db_utils.js
+++ b/server/utils/db_utils.js
@@ -13,14 +13,16 @@ const dbUtils = {
     log.debug('Initializing db...');
 
     // Migrate
-    return dbUtils.migrate(redis).then(
+    return dbUtils.migrate(redis).then(() => {
       // Then seed
-      Promise.all(seedData.map(app =>
+      const appPromises = seedData.apps.map(app =>
         dbUtils.seedApp(redis, dbConstants.REDIS_APPS, app)
-      ))
-    ).then(() => {
+      );
+      const versionPromise = dbUtils.seedVersion(redis);
+      return Promise.all([...appPromises, versionPromise]);
+    }).then(() => {
       log.debug('Initialized db.');
-    }).catch(log.error);
+    });
   },
 
   /**
@@ -39,12 +41,12 @@ const dbUtils = {
         ]).then(() =>
           redis.set(dbConstants.REDIS_VERSION, 1).then(() => {
             log.debug('DB migrated to v1.');
-          }).catch(log.error)
-        ).catch(log.error);
+          })
+        );
       }
 
       return Promise.resolve();
-    }).catch(log.error);
+    });
   },
 
   /**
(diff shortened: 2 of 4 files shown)


server/bin/www
```diff
@@ -5,7 +5,7 @@
  */
 if (!('toJSON' in Error.prototype)) {
   Object.defineProperty(Error.prototype, 'toJSON', {
-    value: () => {
+    value: function toJSON() {
```
justinmc: The arrow function was injecting the parent context into itself, so `this` below didn't reference the error.  Using a regular function fixed the bug and allowed toJSON to give back meaningful stuff.

server/utils/db_utils.js
```diff
@@ -64,7 +66,27 @@ const dbUtils = {
...
+      });
+    });
+  },
+
+  /**
+   * Set the db version to the latest value
+   * @param {RedisClient} redis
+   * @returns {Promise}
+   */
+  seedVersion(redis) {
```
justinmc: I wasn't explicitly seeding the db version number in some cases, so this fixes that.

server/utils/migrations_utils.js
```diff
@@ -7,52 +8,58 @@ const migrationsUtils = {
    * Rename an old field in each run to the new name
    */
   migrateRunsWorkflowField(redis) {
-    global.log.debug('Migrating runs workflow fields...');
```
justinmc: This was the first bug in the migrations.  We moved from a global variable to a `log` module but I guess I forgot to change out these ones.

server/utils/migrations_utils.js
```diff
@@ -7,52 +8,58 @@ const migrationsUtils = {
...
 
           return redis.hset(dbConstants.REDIS_APPS, oldApp.id, JSON.stringify(oldApp)).then(() => {
-            global.log.debug(`Finished migrating workflow ${oldApp.id} to app.`);
-          }).catch(global.log.error);
-        }).catch(global.log.error);
+            log.debug(`Finished migrating workflow ${oldApp.id} to app.`);
+          });
+        });
       });
-    }).catch(global.log.error);
```
justinmc: I removed all these intermediate error handling things like we discussed and instead handle everything at the root level in bin/www where `dbUtils.inititalize` is called.

server/utils/seed_data.js
```diff
@@ -1,101 +1,105 @@
...
-    description: 'Calculate the electron binding energy of an' +
-        ' anionic doublet species using DFT',
-    widgets: [
-      {
-        id: widgetsConstants.LOAD,
-        meta: {
-          title: 'Load Molecule',
+const seedData = {
+  version: 1,
+  apps: [
```
justinmc: I changed `seedData` to an object so I can include seed data other than just apps (in this case I needed the version number).

server/utils/migrations_utils.js
```diff
@@ -7,52 +8,58 @@ const migrationsUtils = {
...
    * Move workflows from the old workflow table into the app table
    */
   migrateWorkflowsToApps(redis) {
-    global.log.debug('Migrating workflows to apps...');
+    log.debug('Migrating workflows to apps...');
 
     return redis.hgetall(OLD_REDIS_APPS).then((oldAppsHash) => {
+      if (!oldAppsHash) {
+        log.debug('Not migrating workflows to apps because no workflows found');
+        return Promise.resolve();
```
justinmc: The other main bug in migrations was that the migrations would choke if there were no old `workflows` to migrate, such as in a fresh build!

justinmc:
@dionjwa Take a look at my comments above about the main points.  This appears to fix seeding/migrating a fresh db for me, and it also handles errors in that place much better.

server/bin/www
dionjwa: Yeah, I've hit this issue before. I didn't realize that arrow functions work differently to named functions .

server/bin/www
```diff
@@ -110,7 +110,7 @@ dbUtils.initialize(redis).then(() => {
   server.on('listening', onListening(server.address()));
   server.timeout = 8 * 60 * 1000; // 8 minutes
 }).catch((err) => {
-  log.error({ error: err });
+  log.error({ error: err.toJSON() });
 });
```
dionjwa: Shouldn't this catch be removed completely? I think the server should crash if there is an unresolved error at this state in the app.

justinmc: Whoops I forgot about that.  I'll add the promise exception event handler and then remove this.

PR #271, opened 2017-04-05
Working app session API

### LINKED ISSUE #246: Email address "auth"
Switch to a more "auth"-like process, where users enter their email _before_ using the app, not as the last step

dionjwa:
Fixes #246

 - Refactored server app to use a global config object that provides various services via promises, removing the problem of circular dependencies and initialization order.
 - Added API for app sessions (see README.md for API docs).
 - A websocket is used to receive app session updates.
 - Uses SQL database for storing session data. 
 - Includes functional tests for starting an app session, making changes, and getting results via the websocket.


server/main/config.js
```diff
@@ -0,0 +1,41 @@
...
+ * redis pub/sub channels, websocket
+ */
+const Promise = require('bluebird');
+const redisConnectionCreator = require('./redis_connection');
+/*
+ * The global config object containing the various services.
+ * All objects are promises, deferring the creation
+ * of the value to a later time. This avoids issues
+ * around circular imports, and too many direct
+ * require statements to hard-coded relative paths.
```
justinmc: I don't think I can talk you out of it, but I wouldn't use a big singleton global variable like this.  Architecture has worked out a lot better for me when I clearly defined my dependencies on simple modules with no circular dependencies, even if I had to type a few extra `../`.

At least I'd lobby to remove the use of `global` itself and just import the `config` module in those places.  In `bin/www` it's being referenced in both ways confusingly.

server/package.json
```diff
@@ -31,19 +37,36 @@
     "molecular-design-applications-shared": "../shared",
     "morgan": "^1.7.0",
     "mustache": "^2.3.0",
+    "pg": "^6.1.5",
+    "pg-hstore": "^2.3.2",
```
dionjwa: They're not strict dependencies of `sequelize` yet they're needed for postgresql: http://docs.sequelizejs.com/en/v3/docs/getting-started/

It's a bit annoying, but I can see why they make those packages optional.

justinmc: Ah got it.

server/package.json
```diff
@@ -17,11 +21,13 @@
     "bluebird": "^3.4.6",
     "bluebird-retry": "^0.10.1",
     "body-parser": "^1.15.2",
+    "bufferutil": "^3.0.0",
```
dionjwa: It's an optional dependency of ws to make it faster:
https://www.npmjs.com/package/ws

server/package.json
```diff
@@ -31,19 +37,36 @@
...
+    "pg-hstore": "^2.3.2",
     "promise-redis": "0.0.5",
     "redis": "^2.6.3",
     "request": "^2.79.0",
     "request-promise": "^4.1.1",
     "sendgrid": "^4.7.1",
+    "sequelize": "^3.30.4",
     "shortid": "^2.2.6",
-    "validator": "^6.2.0"
+    "utf-8-validate": "^3.0.1",
```
dionjwa: Same as the buffer-util package: https://www.npmjs.com/package/ws

server/routes/app.js
```diff
@@ -30,45 +29,49 @@ router.get('/exitcode/:runId', (req, res) => {
...
-  redis.hget(dbConstants.REDIS_APPS, appId).then((appString) => {
-    if (!appString) {
-      const error = new Error(
-        `No app found for given app id ${appId}`
-      );
-      error.status = 404;
-      return next(error);
-    }
+  global.config.redis()
+    .then(redis => redis.hget(dbConstants.REDIS_APPS, appId))
```
dionjwa: I think short term we store them as static json, versioned, and have the server instances simply load them into memory. It's safe, performant, versioned, and functional tests can be performed. Long term however, would be an sql database, backed by a catalogue or gallery *separate* from the main server instance. Until we get there, I think we should just essentially statically link them. But I'm not addressing that in this PR, the app sessions have no idea about the underlying apps, it's just an id to the app session API.

server/routes/app.js
```diff
@@ -30,45 +29,49 @@ router.get('/exitcode/:runId', (req, res) => {
...
-  Promise.all([
-    redis.hgetall(dbConstants.REDIS_APPS),
-    redis.hgetall(dbConstants.REDIS_RUNS),
-  ]).then(([appsHash, runsHash]) => {
-    const runCounts = appUtils.getRunCountsByApps(runsHash || {});
+  global.config.redis()
+    .then(redis =>
+      Promise.all([
+        redis.hgetall(dbConstants.REDIS_APPS),
+        redis.hgetall(dbConstants.REDIS_RUNS)])
```
dionjwa: Yes, absolutely, but that's not in the scope of this PR. For a while, we're going to have dual (therefore perhaps confusing) modes until we've got everything in.

server/routes/app.js
justinmc: Got it, I'll start moving in that direction as we go forward.

server/routes/app.js
justinmc: 👍

server/routes/app.js
```diff
@@ -30,45 +29,49 @@ router.get('/exitcode/:runId', (req, res) => {
...
 /**
  * Get all apps, including their run count
  */
 router.get('/', (req, res, next) => {
-  Promise.all([
-    redis.hgetall(dbConstants.REDIS_APPS),
-    redis.hgetall(dbConstants.REDIS_RUNS),
-  ]).then(([appsHash, runsHash]) => {
-    const runCounts = appUtils.getRunCountsByApps(runsHash || {});
+  global.config.redis()
```
dionjwa: Fixed. global.config.redis is a promise (all config values are promises for consistency).

server/routes/workflow.js
```diff
@@ -0,0 +1,85 @@
+/**
+ * workflow routes
```
dionjwa: Mistake from rebasing, fixed.

server/services/appsession.js
```diff
@@ -0,0 +1,193 @@
+/**
```
dionjwa: Done.

server/routes/run.js
```diff
@@ -3,42 +3,47 @@ const isEmail = require('validator').isEmail;
 const statusConstants = require('molecular-design-applications-shared').statusConstants;
 const dbConstants = require('../constants/db_constants');
 const log = require('../utils/log');
-const redis = require('../utils/redis');
 const runUtils = require('../utils/run_utils');
 
 const router = new express.Router();
 
+function getRedis() {
+  return global.config.redis;
```
dionjwa: Changed, that was an oversight missed when refactoring.

server/bin/www
```diff
@@ -147,41 +151,58 @@ function onListening(addr) {
...
+    server.timeout = 8 * 60 * 1000; // 8 minutes
+    config.server.resolve(server);
+  })
+  .then(() => new WebsocketServer({ config }))
+  .then(() => {
+    /* Create all the remaining services */
+    global.config.db.resolve(db);
+    global.config.session.resolve(new AppSession({ config }));
+    global.config.wsHandler.resolve(new WebsocketHandler({ config }));
+    global.config.notifications.resolve(new Notifications({ config }));
```
dionjwa: The problem with this approach is that the server is not immediately responsive, so anything that wants to hit the server, e.g. version checks or tests, will have to wait some unknown time.  This way, you can add more services and not have to worry about the order, and you can immediately respond to requests. 

However, if you feel strongly about it, I'll modify it so that those objects are not promises but the actual services, but will involve ensuring the correct order of initialization (which isn't necessarily a bad thing).

dionjwa:
Ok @justinmc made all requested changes.

server/main/config.js
dionjwa: I've also refactored to not use promises in the config object.

justinmc: Ok this looks good to me, thanks for taking care of all my complaints.  I like the new setup where we explicitly write out the initialization order, and then everywhere else in the code we know that all services available.  Hopefully that will make things simpler.

I think I can get behind the dependency injection stuff too.  Especially in these cases, the injected objects are stateful, which probably makes more sense as a function parameter than a required module.

Issue #251, opened 2017-03-28
Decouple app definitions from the server code


avirshup:
Clearly separate the app definitions from the rest of the MST user-facing server

dionjwa:
That means putting the app definitions in a separate git repo that we either load directly from the servers, or put into a db (I favour the former until we need actual app modifications outside the deployment process, and git automatically gives us versioning).

dionjwa:
Currently we load the workflows/apps into redis on each server startup, then load the apps back. This doesn't make much sense to me. For simplicity, I think we should commit the repo of apps as a submodule to the mst server code, so we can control the versioning directly, and have functional tests running on the correctly versioned apps.

PR #273, opened 2017-04-05
Separate app json

justinmc:
Issue https://github.com/Autodesk/molecular-simulation-tools/issues/251

~~Waiting on https://github.com/Autodesk/molecular-simulation-tools/pull/265~~

This PR will move app data into a separate location, remove all redis app data (including the dynamic `viewCount`), and reference apps directly from the static json.

It was mentioned that we might move the static json into a separate git repository.  I'm currently not doing this as a part of this PR since it's a long process, but the folder where the apps json is could easily be moved into its own repo if it's necessary.


justinmc:
@dionjwa Let me know what you think about how I did the app json separation when you're back.  I moved the app json into its own module in a root folder in the project.  If we want to go through the process of creating a new open source repo, it should be very easy to move the folder there.  For now I'm referring to the version of the json as specified in package.json, but if we move to a separate repo and wanted to specify git sha instead, that would also be just as easy.

Notice that I also got rid of all app data in redis and instead refer to apps directly from their static json on the server.  The `viewCount` that was displayed on the homepage has been removed.

Lastly, Travis is giving me the error "TIRRENU_URL not specified", is that due to the same problem that Day was having?

apps-data/apps/vde.js
```diff
@@ -0,0 +1,55 @@
...
+      inputs: [
+        { id: 'final_structure.pdb' },
+        { id: 'results.json' },
+        { id: 'minstep.0.pdb' },
+        { id: 'minstep.1.pdb' },
+        { id: 'minstep_frames.json' },
+      ],
+    },
+  ],
+};
```
dionjwa: Wouldn't it make more sense for these to be actual JSON files that are loaded and exported? Eventually if they live in a database, they're going to be consumed as plain JSON/yaml right? In addition, it makes is slightly harder for others who aren't javascript developers to know if this is JSON. It's obvious to us, but if it is plain JSON, it is easier for others to know we are creating and consuming workflows defined in JSON.

dionjwa: Oh I see why, you are using shared constants. Hmmm.

server/routes/app.js
```diff
@@ -30,45 +31,32 @@ router.get('/exitcode/:runId', (req, res) => {
...
 router.get('/', (req, res, next) => {
-  Promise.all([
-    config.redis.hgetall(dbConstants.REDIS_APPS),
-    config.redis.hgetall(dbConstants.REDIS_RUNS)]
-  )
-  .then(([appsHash, runsHash]) => {
-    const runCounts = appUtils.getRunCountsByApps(runsHash || {});
+  config.redis.hgetall(dbConstants.REDIS_RUNS)
+    .then((runsHash) => {
+      const runCounts = appUtils.getRunCountsByApps(runsHash || {});
```
dionjwa: Are we still needing the REDIS_RUNS?

server/utils/db_utils.js
```diff
@@ -13,14 +13,9 @@ const dbUtils = {
...
-      // Then seed
-      const appPromises = seedData.apps.map(app =>
-        dbUtils.seedApp(redis, dbConstants.REDIS_APPS, app)
-      );
-      const versionPromise = dbUtils.seedVersion(redis);
-      return Promise.all([...appPromises, versionPromise]);
-    }).then(() => {
+    return dbUtils.migrate(redis).then(() =>
+      dbUtils.seedVersion(redis)
+    ).then(() => {
```
dionjwa: Can we write promises like
```
return dbUtils.migrate(redis)
   .then(() => dbUtils.seedVersion(redis))
   .then(() => {...
```
I find it makes the promise steps easier to parse, and also easier to add/subtract steps or error handers when needed.

apps-data/apps/vde.js
justinmc: Yeah exactly.  I agree plain json will probably be easier for people though.  I guess I'll leave it for now and if we start on boarding app developers that want to have json, then I can replace all the constants with repeated strings.

server/routes/app.js
justinmc: For now yes, but I'm working on a PR now to move everything to your sessions API.

server/utils/db_utils.js
justinmc: Yup sorry, please keep reminding me until I remember this haha.  I wish there was a linter rule for it.

Dockerfile
```diff
@@ -74,6 +74,12 @@ RUN npm run build
 #######################################
 # Server build/install packages
 #######################################
+RUN mkdir -p $APP/apps-data
+ADD ./apps-data $APP/apps-data
+WORKDIR $APP/apps-data
+RUN npm install
+WORKDIR $APP
+
```
justinmc: Done, thanks for the tips.

PR #275, opened 2017-04-07
Working CCC/turbo API routes, with functional tests

dionjwa:


README.md
```diff
@@ -120,6 +120,100 @@ Returns the session state
...
+  "stats": {
+    "copyInputs": "0.149s",
+    "ensureImage": "0s",
+    "containerCreation": "0.161s",
+    "containerExecution": "1.302s",
+    "copyOutputs": "0.208s",
+    "copyLogs": "0.003s",
+    "total": "1.5110000000000001s"
+  }
+}
```
justinmc: Did you mean to indent this to get the codeblock?  Same question for the other "Returns" block below.

server/main/db.js
```diff
@@ -3,12 +3,16 @@
...
 const log = require('../utils/log');
 
 assert(process.env.DB_CONNECTION_URI, 'Missing env var DB_CONNECTION_URI');
 const db = new Sequelize(process.env.DB_CONNECTION_URI);
 
-module.exports = db.authenticate()
+module.exports = retry(
+    () => db.authenticate(),
+    { max_tries: 50, interval: 1000, max_interval: 10000 }
+  )
```
justinmc: Good call on retrying this.

server/routes/ccc.js
```diff
@@ -0,0 +1,48 @@
...
+  config.ccc.runTurbo(req.body)
+    .then(result => res.json(result))
+    .catch(next);
+});
+
+router.post('/run/:sessionId/:widgetId', (req, res, next) => {
+  const sessionId = req.params.sessionId;
+  const widgetId = req.params.widgetId;
+  assert(sessionId, 'Missing sessionId');
+  assert(widgetId, 'Missing widgetId');
```
justinmc: Can you get these asserts to ever throw an error?  I've wondered this about my own code too.  If you don't pass `sessionId` or `widgetId` in the URL, then you won't hit this route at all.

server/services/ccc.js
```diff
@@ -0,0 +1,194 @@
...
+    }
+ */
+CCC.prototype.runTurbo = function runTurbo(jobBlob) {
+  return this.ccc
+    .then((ccc) => {
+      const promises = [];
+      const inputs = {};
+      if (jobBlob.inputs) {
+        Object.keys(jobBlob.inputs).forEach((key) => {
+          const inputBlob = jobBlob.inputs[key];
```
justinmc: FYI I was playing around with Object.entries recently, and you can also do this:

`Object.entries(jobBlob.inputs).forEach(([key, inputBlob]) => {`

server/services/ccc.js
```diff
@@ -0,0 +1,194 @@
...
+          if (inputBlob.type === 'url') {
+            promises.push(
+              request(inputBlob.value)
+                .then((result) => {
+                  inputs[key] = result;
```
justinmc: Will these inputs ever be too big to send over the network to CCC?

server/services/ccc.js
```diff
@@ -0,0 +1,194 @@
...
+      return Promise.all(promises)
+        .then(() => {
+          jobBlob.inputs = inputs;
+          return ccc.submitTurboJobJson(jobBlob);
+        });
+    });
+};
+
+/**
+ * [turboJob description]
```
justinmc: I think you meant to replace this with a description, and also this one is not a turboJob right?

server/test/test_session.js
```diff
@@ -0,0 +1,179 @@
...
+      })
+      .catch((err) => {
+        log.error(err);
+        return { success: false, error: JSON.stringify(err).substr(0, 500) };
+      });
+  },
+
+};
+
+module.exports = utils;
```
justinmc: Thanks for writing all of these tests for this PR!

server/utils/run_utils.js
```diff
@@ -94,12 +94,12 @@ const runUtils = {
...
-      .then(ccc =>
-        ccc.getJobResult(runId)
-      );
-  },
+  // waitOnJob(runId) {
+  //   return cccUtils.promise()
+  //     .then(ccc =>
+  //       ccc.getJobResult(runId)
+  //     );
+  // },
```
justinmc: Did you mean to leave this commented out?  I guess we should be moving off of this code entirely anyway right?

server/services/ccc.js
```diff
@@ -0,0 +1,194 @@
...
+        .then(() => cccConnection);
+    }
+    return cccConnection;
+  });
+
+  this.jobSessionQueue = Queue('mst:job_session_queue',
+    this.redisConfig.port, this.redisConfig.host);
+
+  this.jobSessionQueue.process((job, done) => {
+    this.processQueue(job, done);
```
justinmc: Awesome use of a queue for robustness.  So if the server restarts in the middle of running some jobs, then when it comes back up it will hit this line here and return to taking care of those jobs?

dionjwa: Basically yes. It's a little more complicated, but the queue is persistent, and handles job processors restarting or getting killed (it will re-add the job to the queue).

server/services/ccc.js
dionjwa: NICE! I'll change it to that. That is much more succinct.

server/services/ccc.js
dionjwa: The server itself has a limit of 50mb, we might hit that limit eventually, but CCC doesn't care.

server/services/ccc.js
dionjwa: Fixed.

server/utils/run_utils.js
dionjwa: Missed that, removed.

PR #279, opened 2017-04-11
Interactive sim

dane1122:
- Updated interactive-sim branch. Much faster simulation since using **nve** simulation instead of **nvt**. 
- Also changed demo molecules so one of them is larger than the other.


dane1122:
**MST app: a6605c1 and 3d958a8**

client/public/js/utils/interactive_molecule_viewer_tools.js
```diff
@@ -0,0 +1,98 @@
+import interactiveSimConstants from '../constants/interactive_sim_constants';
+
+const InteractionTool = {
```
justinmc: This should start with a lower case i by convention: `interactionTool` because it is not a newable class.  Not a big deal though, especially if the linter isn't complaining.

interactive-sim/index.html
```diff
@@ -94,8 +94,7 @@
 				<div class="radio" title="Apply force to selected atoms (RECOMMENDED)">
```
justinmc: Is the plan to remove this static page once the interactive sim app is working?

server/routes/structure.js
```diff
@@ -67,4 +67,22 @@ router.post('/executeWorkflow1Step0', (req, res, next) => {
...
 });
 
+// Workflow for interactive molecular dynamics molecule preparation
+router.post('/executeWorkflow4Step0', (req, res, next) => {
+  const inputs = req.body.inputs;
+  if (!inputs) {
+    return next(new Error('No inputs'));
+  }
+
+  return workflowUtils.executeWorkflow1Step0(inputs)
```
justinmc: Do you plan to add Aaron's PDB validation in another PR later?

interactive-sim/index.html
dane1122: I think I would need to talk to Flo+Aaron for that!

server/routes/structure.js
dane1122: I think Aaron will create a brand new workflow for my app since it requires some work to be done with the molecule!

interactive-sim/index.html
justinmc: 👍

server/routes/structure.js
justinmc: Sounds good

PR #293, opened 2017-05-19
285 generic execution calls v1

dionjwa:
Partially complete actions to call the various ccc compute methods, and to establish the websocket connection to the server.

### DIFF
--- a/apps-data/apps/vde.js
+++ b/apps-data/apps/vde.js
@@ -2,33 +2,71 @@ const widgetsConstants = require('molecular-design-applications-shared').widgets
 
 module.exports = {
   title: 'Calculate electronic vertical detachment energy',
-  selectLigands: false,
-  bgColor: '#292E60',
-  bgIndex: 3,
-  color: '#2FE695',
-  comingSoon: false,
-  creatorImage: '/img/logo2.png',
-  description: 'Calculate the electron binding energy of an' +
-      ' anionic doublet species using DFT',
+  version: "0.1-alpha",
+  meta: {
+    name: "VDE",
+    version: "0.0.1",
+    description: "Calculate the electron binding energy of an anionic doublet species using DFT",
+    git: "https://github.com/Autodesk/molecular-simulation-tools",
+    authors: [],
+    display: {
+      selectLigands: false,
+      bgColor: '#292E60',
+      bgIndex: 3,
+      color: '#2FE695',
+      comingSoon: false,
+      creatorImage: '/img/logo2.png',
+    }
+  },
   widgets: [
     {
-      id: widgetsConstants.LOAD,
-      type: widgetsConstants.LOAD,
-      title: 'Load Molecule',
-      outputs: [
+      id: "load_pdb",
+      type: widgetsConstants.RUN_DOCKER_CONTAINER_FAST,
+      meta: {
+        title: 'Load Molecule',
+      },
+      config: {
+        image: "avirshup/mst:workflows-0.0.alpha5",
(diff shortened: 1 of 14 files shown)


client/public/js/actions.js
```diff
@@ -428,3 +429,73 @@ export function changeMorph(morph) {
...
+        console.error(err);
+        dispatch({
+          type: actionConstants.CCC_RUN_ERROR,
+          sessionId,
+          widgetId,
+          error: err
+        });
+      });
+  };
+}
```
justinmc: These actions look pretty good to me.  Right now, clicking the Run button is calling the action [clickRun](https://github.com/Autodesk/molecular-simulation-tools/blob/master/client/public/js/actions.js#L122).  I've filled it out with some fake data for now so we can work on the rest of the app.  That one will have to get merged with these in one way or another.

client/public/js/components/app_router.jsx
```diff
@@ -18,6 +18,47 @@ class AppRouter extends React.Component {
...
+          break;
+      }
+    });
+  }
+
+  componentWillUnmount() {
+    if (this.ws) {
+      this.ws.close();
+      this.ws = null;
+    }
```
justinmc: This component will unmount when the user changes from routes with an appId in them to any other route (see [main.jsx](https://github.com/Autodesk/molecular-simulation-tools/blob/master/client/public/js/main.jsx) for the full routes).  However, if you switch directly from one session to another or one app to another, this app will not unmount and the websocket will remain.

I think it's going to be tough to find get a component's lifecycle to coincide exactly with the websocket lifecycle we want.  It looks to me like it would be better off keeping the reference in a store.

server/main/db.js
```diff
@@ -7,34 +6,34 @@ const retry = require('bluebird-retry');
...
   config.database = process.env.PGDATABASE;
 }
 
-console.log('Connecting to database with settings: \n database=' + config.database + '\n username=' + config.username + 
-            '\n password=***removed**' + '\n host=' + config.host  + '\n port=' + config.port);
+console.log(`Connecting to database with settings: \n database=${config.database} \n username=${config.username} 
+            \n password=***removed**\n host=${config.host}\n port=${config.port}`);
 
-let db = new Sequelize(config.database, config.username, config.password, {
+const db = new Sequelize(config.database, config.username, config.password, {
```
justinmc: Thanks for cleaning this up!

client/public/js/components/widget_ccc_turbo.jsx
```diff
@@ -0,0 +1,15 @@
...
+      {props.children}
+    </div>
+  );
+}
+
+WidgetCCCTurbo.propTypes = {
+  children: React.PropTypes.element,
+};
+
+export default WidgetCCCTurbo;
```
justinmc: It might be a better idea just to reuse the `status_run.jsx` component for now.

apps-data/apps/vde.js
```diff
@@ -2,33 +2,71 @@ const widgetsConstants = require('molecular-design-applications-shared').widgets
...
   title: 'Calculate electronic vertical detachment energy',
-  selectLigands: false,
-  bgColor: '#292E60',
-  bgIndex: 3,
-  color: '#2FE695',
-  comingSoon: false,
-  creatorImage: '/img/logo2.png',
-  description: 'Calculate the electron binding energy of an' +
-      ' anionic doublet species using DFT',
+  version: "0.1-alpha",
```
dionjwa: Ugh. Yeah. I started moving this to basic JSON.

PR #294, opened 2017-05-22
Progress: loading generic call working

dionjwa:


apps-data/apps/vde.js
```diff
@@ -44,11 +52,36 @@ module.exports = {
...
+          source: {
+            id: "clean_pdb", pipe: "minstep.1.pdb"
+          }
+        },
+        {
+          name: "minstep_frames.json",
+          source: {
+            id: "clean_pdb", pipe: "minstep_frames.json"
+          }
+        }
```
justinmc: Making a note that we should modify the other app(s) to use this new format as well.  The new input format and the config object above.

client/public/js/actions.js
```diff
@@ -130,79 +130,24 @@ export function clickRun(appId, widget, runId, email, pipeDatasByWidget, inputSt
...
+    console.log(`clickRun inputPipeDatas=${inputPipeDatas}`);
+
+    apiUtils.runCCC(runId, widget.id, widget.config, inputPipeDatas)
+      .then((cccResult) => {
+        console.log('cccResult', cccResult);
+        // dispatch({
+        //   type: actionConstants.RUN_SUBMITTED,
+        //   runId,
+        //   widgetId: widget.id,
+        // });
```
justinmc: TODO: Update the session data to include the jobId (on the server and the client).

client/public/js/actions.js
```diff
@@ -420,3 +365,61 @@ export function changeMorph(morph) {
     morph,
   };
 }
+
+export function runCCC(runId, widget, inputMap) {
```
justinmc: I guess this is the generic version of `clickRun`.

client/public/js/actions.js
```diff
@@ -420,3 +365,61 @@ export function changeMorph(morph) {
...
+          type: actionConstants.CCC_RUN_ERROR,
+          runId,
+          widget,
+          error: err,
+        });
+      });
+  };
+}
+
+export function updatePipeData(runId, pipeDatasByWidget) {
```
justinmc: I worry about moving the logic that updates pipeDatas into the components.  It might make sense for long term generic widget components though.  It would be a big refactor to get our existing components to use this for now.

client/public/js/components/app.jsx
```diff
@@ -55,12 +63,108 @@ class App extends React.Component {
...
   initialize(appId, runId) {
     if (runId) {
+      this.initializeWebsocket(runId);
       return this.props.initializeRun(appId, runId);
     }
 
     return this.props.initializeApp(appId);
   }
 
+  initializeWebsocket(runId) {
```
justinmc: I think we should move this into actions.js and use a closure variable for the websocket.  As it is, the component lifecycle won't sync up exactly with the websocket lifecycle as we want it.  Navigating directly from one app to another will not close the old app's web socket.

client/public/js/utils/api_utils.js
```diff
@@ -172,27 +154,43 @@ const apiUtils = {
...
-    return axios.post(`${API_URL}/v1/structure/executeApp${appId}Step0`, data)
+    jobData.inputsPath = '/inputs';
+
+    for (let i = 0; i < jobData.command.length; i += 1) {
+      if (jobData.command[i].indexOf('input.pdb')) {
+        jobData.command[i] = jobData.command[i].replace('input.pdb', `input.${nameExtension}`);
+      }
+    }
+
+    return axios.post(`${API_URL}/v1/ccc/run/turbo`, jobData)
```
justinmc: I'd move this to api_utils.js just for consistency.

apps-data/apps/vde.js
justinmc: Actually this is probably something we should put off until another PR since it requires a lot of refactoring and it mostly separate from what we're working on.  Is this so that you can have multiple inputs from the same source?

client/public/js/components/app.jsx
dionjwa: Are you sure? If the app id or run id changes, the websocket is closed, and then recreated if there is a valid runId.

justinmc: If the app id changes then the component doesn't unmount, so `ws.close` isn't called.  It might be possible to hack this functionality into the component, but I don't see why we would when the actions is where all the rest of the logic like this has been, and it also gives us much more control over when to close the websocket.

I don't want to create more work for this MVP push but I think we should keep in mind that it should be moved to the actions.

apps-data/apps/vde.js
dionjwa: It's mostly so there's no assumption about the connections between widgets.

client/public/js/components/app.jsx
dionjwa: The websocket is also update on the initialize call, which is triggered on an app or run id change.

dionjwa: Not only the unmount.

dionjwa: I'm very against keeping a websocket in the state or props. It shouldn't be difficult to control its lifecycle with a combination of runId/AppId and mount/unmount events.

justinmc: I see how it would be possible, but I don't think your logic is covering all cases.  Try entering your email to start a new session, then press the back button to go back to the empty app.  It looks to me like the websocket wouldn't get closed in that case.

Also I'm suggesting storing it in a closure variable in the actions, not in props or state.  I'm not going to refactor anything right now but I'll try to keep an eye out for a way to handle this asynchronous logic somewhere that's more in line with how all the other asynchronous logic is handled in this app.  Maybe even doing some of this in a middleware might end up being nice.

PR #295, opened 2017-05-23
Run Widget Status

justinmc:
This PR gets rid of the global thankyou/error pages and moves them into states of the StatusRun component.  Run data is currently mocked since that will come from the new api endpoints soon.


client/public/js/components/app.jsx
```diff
@@ -1,96 +1,177 @@
...
+    );
+  }
 }
 
+App.defaultProps = {
+  runId: '',
+};
+
 App.propTypes = {
+  app: React.PropTypes.instanceOf(AppRecord).isRequired,
```
dionjwa: If you are passing in the app, why also have the app.id?

justinmc: @dionjwa `appId` is coming from the URL, which is available right away, while `app` has to be downloaded based on that appId.  This component handles calling `initializeApp` to download the app when needed, so it needs both.

PR #297, opened 2017-05-25
Session Test Fix

justinmc:
Fix the broken unit tests.


client/public/js/containers/app_root.jsx
```diff
@@ -95,6 +95,7 @@ function mergeProps(stateProps, dispatchProps) {
     clickRun: dispatchProps.clickRun(
       stateProps.app.run.id,
       stateProps.app.run.email,
+      stateProps.app.run.pipeDatasByWidget,
```
justinmc: Just noticed that this data wasn't being piped in where it was expected so I hooked it up.  Didn't appear to be related to any unit tests failing.

client/public/js/utils/app_utils.js
```diff
@@ -56,6 +56,10 @@ const appUtils = {
     return Promise.all(pipeDatas.map((pipeData) => {
       if (!pipeData.pipeName.endsWith('.json')) {
         return Promise.resolve(pipeData);
+      } else if (pipeData.type === 'inline') {
+        return Promise.resolve(
+          pipeData.set('fetchedValue', JSON.parse(pipeData.value)),
+        );
```
justinmc: Making sure that inline json components have their parsed json set on `fetchedValue`.

client/test/components/app_spec.jsx
```diff
@@ -2,16 +2,15 @@ import React from 'react';
 import sinon from 'sinon';
```
justinmc: This file just needed to be renamed after I renamed the corresponding component in another PR and forget about its unit test.

client/public/js/containers/app_root.jsx
dionjwa: Thanks, yeah I also fixed that locally. I think we can remove the email argument pretty soon.

client/public/js/utils/app_utils.js
dionjwa: Is the JSON.parse function going to mess up the expected data type later? For example, what if pipeData.value is a string, and meant to be a string?

client/test/utils/pipe_utils_spec.js
```diff
@@ -135,12 +135,12 @@ describe('pipeUtils', () => {
...
-    describe('when prep.json with no fetchedValue', () => {
+    describe('when prep.json with no value', () => {
       beforeEach(() => {
         inputPipeDatas = inputPipeDatas.push(new PipeDataRecord({
           pipeName: 'prep.json',
-          type: 'url',
-          value: 'http://localhost:9000/r17IbGbKg/outputs/prep.json',
+          type: 'inline',
+          value: '',
         }));
```
dionjwa: In general, value's should be allowed to be null. I'm uncomfortable for an empty string being a default value, since it's not null.

client/public/js/utils/app_utils.js
justinmc: This code will only run if `pipeName` ends in ".json", so at least we shouldn't have any problems with it in our MVP apps.  I agree we should do more to support plain strings for the "inline" type and other formats.  Right now it would probably break somewhere else if not here.

client/test/utils/pipe_utils_spec.js
justinmc: I'm on board with that, I'll try to remember to keep this null by default going forward.

PR #298, opened 2017-05-26
285 generic execution calls v2

dionjwa:
This PR is a WIP.

State: loading molecule working, but the new run widget needs proper data in/out.

### DIFF
--- a/client/public/js/actions.js
+++ b/client/public/js/actions.js
@@ -120,12 +120,17 @@ export function clickWidget(widgetIndex) {
  * @param {String} [inputString]
  */
 export function clickRun(widget, runId, email, pipeDatasByWidget) {
+  console.log('widget', widget);
+  console.log('runId', runId);
+  console.log('email', email);
+  console.log('pipeDatasByWidget', pipeDatasByWidget);
   return async function clickRunAsync(dispatch) {
     dispatch({
       type: actionConstants.CLICK_RUN,
       widgetId: widget.id,
     });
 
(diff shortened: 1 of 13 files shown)


client/public/js/actions.js
```diff
@@ -242,7 +247,15 @@ export function submitInputString(inputString, widget, runId, pipeDatasByWidget)
         );
       });
 
-      await apiUtils.updateSession(runId, updatedPipeDatasByWidget); /* eslint no-unused-expressions: 'off', max-len: 'off' */
+      module.exports.updateWidgetPipeData(runId, widget.id, inputPipeDatas);
```
justinmc: You can just reorder these so that `updateWidgetPipeData` is defined above and then use it directly.

client/public/js/components/app.jsx
```diff
@@ -135,27 +137,36 @@ class App extends React.Component {
...
-              updatedPipeDatasByWidget = updatedPipeDatasByWidget.set(widgetId, widgetPipeDataList);
-              Object.keys(widgetBlob.out).forEach((outPipeName) => {
-                const pipeBlob = widgetBlob.out[outPipeName];
-                const pipeDataRecord = new PipeDataRecord({
-                  pipeName: outPipeName,
-                  type: pipeBlob.type,
-                  value: pipeBlob.value,
-                  widgetId,
+
+            // Only update the session data if there's an email
```
justinmc: I see what you mean now

Issue #300, opened 2017-06-02
Replace "bionano" with "lifesciences" for links, emails, etc


dionjwa:
![image](https://cloud.githubusercontent.com/assets/260221/26736955/04ea7c50-477d-11e7-9e6a-7790f25e052c.png)

PeterRJones:
Team name in email text should also be changed. `BioNano Research Group` to `Lifesciences Group`

PR #303, opened 2017-06-02
Rename bionano to Life Sciences

justinmc:
Issue https://github.com/Autodesk/molecular-simulation-tools/issues/300

Some notable things that I DID NOT change:

  - Link to our blog, which still seems to include `bionano` for the near future.
  - URLs to several internal services that seem to still include `bionano`

@dionjwa You should probably run `ag bionano` and just do a double check that none of those services I mentioned should have their names changed.  They all look like they'd never be exposed to the user to me, though.


dionjwa:
Looks good. Yeah all the others are external or internal services that still are using bionano for the foreseeable future (quay.io/blog/aws internal DNS entries).

